PR #13, opened 2016-07-13
Remove unwrap calls in description parsing.

masklinn:
The unwrap calls are pre-checked and safe, but they're also unnecessary.


kinetiknz:
Looks good, thanks!

PR #14, opened 2016-07-27
Default impls and documentation ticks.

leodasvacas:
Implement `Default` for some types and add some ticks to documentation.


kinetiknz:
Thanks!

PR #16, opened 2016-08-27
Set some structs/fields public

### LINKED ISSUE #15: How to get info?
After using the `read_mp4` function, what am I supposed to do with the `MediaContext` struct since all its fields are hidden? And this is despite the fact the doc says:

> Metadata is accumulated in the passed-through MediaContext struct, which can be examined later.

GuillaumeGomez:
Fixes #15.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mp4parse"
-version = "0.4.0"
+version = "0.4.1"
 authors = [
   "Ralph Giles <giles@mozilla.com>",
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Use in [video-metadata-rs](https://github.com/GuillaumeGomez/video-metadata-rs/pull/24).

rillian:
Seems reasonable. We've been doing everything through the capi. Do you want to add something to `tests/` to confirm these symbols remain available?

GuillaumeGomez:
Sure! I'll add tests.

PR #17, opened 2016-08-29
Bug 1240412: add mp4parse_is_fragmented() for fragmented video checking

alfredoyang:


alfredoyang:
I'm not sure the test should be in gecko mochitest or rust test.

rillian:
Right now gecko doesn't run the rust tests, although I hope it will soon. So for small things it's good to have coverage in both places. Things that require large test files are probably better in gecko to keep this repo small.

src/capi.rs
```diff
@@ -444,6 +444,26 @@ pub unsafe extern fn mp4parse_get_track_video_info(parser: *mut mp4parse_parser,
     MP4PARSE_OK
 }
 
+#[no_mangle]
+pub unsafe extern fn mp4parse_is_fragmented(parser: *mut mp4parse_parser, track_id: u32, fragmented: *mut bool) -> mp4parse_error {
```
rillian: The extern api is C, not C++ so it is safer to use `*mut u8` instead of `*mut bool` for the out-parameter here. I'm sure it works, because rust-llvm and C stdbool will likely end up with the same representation, but this isn't guaranteed.

src/capi.rs
```diff
@@ -444,6 +444,26 @@ pub unsafe extern fn mp4parse_get_track_video_info(parser: *mut mp4parse_parser,
...
+    if parser.is_null() || (*parser).poisoned() {
+        return MP4PARSE_ERROR_BADARG;
+    }
+
+    let context = (*parser).context_mut();
+    let tracks = &context.tracks;
+    (*fragmented) = false;
+
+    for track in tracks {
+        if track.track_id.unwrap() == track_id && track.fragmented == true {
```
rillian: Please don't call `unwrap()` in library code. It will crash the application, so this is at least as strong as `MOZ_RELEASE_ASSERT`. Instead check `track.track_id.is_none()` or use a `match` and skip it or return MP4PARSE_ERROR_INVALID depending the behaviour you  want.

Calling `unwrap()` in test code is fine since it's a good way to abort on unexpected state.

src/lib.rs
```diff
@@ -643,7 +644,7 @@ fn read_stbl<T: Read>(f: &mut BMFFBox<T>, track: &mut Track) -> Result<()> {
                 log!("{:?}", stsd);
             }
             BoxType::TimeToSampleBox => {
-                let stts = try!(read_stts(&mut b));
+                let stts = try!(read_stts(&mut b, track));
```
rillian: It feels more natural to me to set `track.fragmented = stts.samples.is_empty()` directly here rather than plumbing track down into `read_stts`. This keeps the read_* functions more uniform, each taking a box and returning just the structure representing that box. But it's also a general guideline in rust to avoid passing mutable state where possible.

rillian:
The patch is fine with the style comments addressed, but could you say something about the approach you took? I would have naively extended `found_moov` with `found_moof` and based the check on that.

rillian:
Oh, I guess we want to know this just from the initialization segment passed through the MediaSource Extensions API, and that contains just the `ftyp` and `moov`. So you're just checking if the moov sample tables contain any data?

rillian:
The [MSE spec](https://w3c.github.io/media-source/isobmff-byte-stream-format.html#iso-init-segments) says we must reject initialization segments which contain non-zero sample counts in `stsc` and `stco`  as well. Should we add the same check to those branches of `read_stbl` as well?

alfredoyang:
Yes, my intention is checking if moov contains any sample table.
I learnt from kinetik that the initial goal of this library is replacing stagefright for moov paring. MoofParser will parse the moof sample table. So I just add an api to check if current video is fragmented or not.
According to spec, stts/stsc/stco are mandatory boxes, I'll check them at next patch.

rillian:
Ok, sounds good.

kinetiknz:
I'd prefer testing for the presence of an mvex box since there's less chance of a false positive.  That's required to be present in fragmented files, and it's used as the test for a fragmented file in other parsers (e.g. Apple's AVFoundation in AVAsset::canContainFragments).

rillian:
I think it's fine to test both, we can always relax later. If you think empty stts/stsc/stco boxes will have false positives, we should ask about amending the spec?

alfredoyang:
After rebase, this pull request looks complicated. Should I create a new one?

rillian:
The merges seem to have confused the github interface. This diff on the command line looks fine. I'd prefer you keep things in this PR so all the discussion is together, but you can rebase and squash everything together unless you particularly want your change history in the main tree.

src/capi.rs
```diff
@@ -444,6 +457,47 @@ pub unsafe extern fn mp4parse_get_track_video_info(parser: *mut mp4parse_parser,
...
+    if parser.is_null() || (*parser).poisoned() {
+        return MP4PARSE_ERROR_BADARG;
+    }
+
+    let context = (*parser).context_mut();
+    let tracks = &context.tracks;
+    (*fragmented) = false as u8;
+
+    if context.has_mvex == false {
+        return MP4PARSE_OK;
```
rillian: Ok, so a track must have an mvex to be fragmented.

src/lib.rs
```diff
@@ -295,33 +296,34 @@ impl Default for TrackType {
...
+    pub empty_duration: Option<MediaScaledTime>,
+    pub media_time: Option<TrackScaledTime>,
+    pub timescale: Option<TrackTimeScale>,
+    pub duration: Option<TrackScaledTime>,
     track_id: Option<u32>,
-    mime_type: String,
-    data: Option<SampleEntry>,
-    tkhd: Option<TrackHeaderBox>, // TODO(kinetik): find a nicer way to export this.
+    pub mime_type: String,
+    fragmented: (bool, bool, bool), // true when there is no data in stts, stsc, and stco.
```
rillian: I guess if you want the presence of any sample table data to prevent a track from being marked fragmented, you need to track all three of these. I'd prefer you used a struct with named fields for that. That makes it easier to read and harder to confuse which flag is which without using extra storage.

You could also use a HashSet with enum values (BoxType would serve well) which is higher overhead but makes checking if they're all set a little nicer.

src/capi.rs
```diff
@@ -444,6 +457,47 @@ pub unsafe extern fn mp4parse_get_track_video_info(parser: *mut mp4parse_parser,
...
+    }
+
+    // check sample tables.
+    for track in tracks {
+        let fragmented_track = match track.track_id {
+            Some(id) => {
+                if id == track_id {
+                    track.fragmented
+                } else {
+                    (false, false, false)
```
rillian: This replaces data from tracks we're not interested in with something expected to fail the subsequent match statement. Can you instead try pulling out the track of interest and testing that directly? Something like

```rust
match tracks.find(|track| track.track_id == track_id) {
    None => return MP4PARSE_ERROR_BADARG,
    Some(track) where track.fragmented == (true, true, true) => (*fragmented) = true,
    _ => (),
}
MP4PARSE_OK
```

You can split that into more lines if you like, the main issue I'd like to avoid is the implicit state passing. There's no reason to test tracks with different track_id's at all.

rillian:
Sorry, #21 bitrotted your changes, but the files have just moved. Feel free to start a new PR if that's easier for you.

alfredoyang:
Ok, I think I found a way to make this pull looks better without those rebase commits.

I use a structure EmptySampleTableBoxes to keep the status of the 3 boxes. And a function EmptySampleTableBoxes::all_empty() to check the condition. I think it'd be look better.

rillian:
Looks good, thanks!

PR #19, opened 2016-08-29
Add test to keep needed fields public

GuillaumeGomez:
As discussed in #16.

### DIFF
--- a/tests/public.rs
+++ b/tests/public.rs
@@ -1,0 +1,89 @@
+/// Check if needed fields are still public.
+
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+extern crate mp4parse as mp4;
+
+use std::io::{Cursor, Read};
+use std::fs::File;
+
+// Took from https://github.com/GuillaumeGomez/audio-video-metadata/blob/9dff40f565af71d5502e03a2e78ae63df95cfd40/src/metadata.rs#L53
+#[test]
+fn check_mp4() {
+    let mut fd = File::open("examples/minimal.mp4").expect("Unknown file");
+    let mut buf = Vec::new();
+    fd.read_to_end(&mut buf).expect("File error");
+
+    let mut c = Cursor::new(&buf);
+    let mut context = mp4::MediaContext::new();
+    mp4::read_mp4(&mut c, &mut context).expect("read_mp4 failed");
+    assert_eq!(context.timescale, Some(mp4::MediaTimeScale(1000)));
+    for track in context.tracks {
+        match track.data {
+            Some(mp4::SampleEntry::Video(v)) => {
+                // track part
+                assert_eq!(track.duration, Some(mp4::TrackScaledTime(512, 0)));
+                assert_eq!(track.empty_duration, Some(mp4::MediaScaledTime(0)));
+                assert_eq!(track.media_time, Some(mp4::TrackScaledTime(0, 0)));
+                assert_eq!(track.timescale, Some(mp4::TrackTimeScale(12800, 0)));
+                assert_eq!(v.width, 320);
+                assert_eq!(v.height, 240);
+
+                // track.tkhd part
+                let tkhd = track.tkhd.unwrap();
+                assert_eq!(tkhd.disabled, false);
+                assert_eq!(tkhd.duration, 40);
+                assert_eq!(tkhd.width, 20971520);
+                assert_eq!(tkhd.height, 15728640);
+
+                // track.data part
+                assert_eq!(match v.codec_specific {
+                    mp4::VideoCodecSpecific::AVCConfig(v) => {
+                        assert!(v.len() > 0);
+                        "AVC"
+                    }
+                    mp4::VideoCodecSpecific::VPxConfig(vpx) => {
+                        // We don't enter in here, we just check if fields are public.
+                        assert!(vpx.bit_depth > 0);
+                        assert!(vpx.color_space > 0);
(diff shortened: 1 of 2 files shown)


tests/public.rs
```diff
@@ -0,0 +1,39 @@
...
+extern crate mp4parse as mp4;
+
+use std::io::Cursor;
+
+// Took from https://github.com/GuillaumeGomez/audio-video-metadata/blob/9dff40f565af71d5502e03a2e78ae63df95cfd40/src/metadata.rs#L53
+//
+// Not important if it fails or not. If it compiles, it means it's all good.
+#[test]
+fn check_mp4() {
+    let mut c = Cursor::new(b"\x00\x00\x00\x13\x66\x74\x79\x70\x30\x30\x30\x30\x30\x30\x30\x30".to_vec());
```
rillian: Theres examples/minimal.mp4 you can test against too.

tests/public.rs
```diff
@@ -0,0 +1,39 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+extern crate mp4parse as mp4;
+
+use std::io::Cursor;
+
+// Took from https://github.com/GuillaumeGomez/audio-video-metadata/blob/9dff40f565af71d5502e03a2e78ae63df95cfd40/src/metadata.rs#L53
+//
+// Not important if it fails or not. If it compiles, it means it's all good.
```
rillian: Surely it should run to completion? :)

tests/public.rs
GuillaumeGomez: Even better, thanks!

rillian:
Thanks. This is a good start, but you marked a lot more fields public. Maybe see how many can be checked with the in-tree test file?

GuillaumeGomez:
Making it more global.

GuillaumeGomez:
Ok, so now we have something even more complete. Also, I added myself as an author, I intend to contribute in here in the future (work linked to my crate). If you don't want it, I'll just remove the commit.

tests/public.rs
```diff
@@ -0,0 +1,101 @@
...
+fn check_mp4() {
+    let buf = if let Some(mut fd) = File::open("examples/minimal.mp4").ok() {
+        let mut buf = Vec::new();
+
+        match fd.read_to_end(&mut buf) {
+            Ok(_) => buf,
+            Err(_) => panic!("File error"),
+        }
+    } else {
+        panic!("Unknown file")
```
rillian: I'd rather have this up near the `File::open` and save a level of indent. For example you could use a pair of `match` stanzas like in `examples/dump.rs`. I think the readability is more important than freeing the file object sooner.

Using .expect() would be appropriate in a test too.

tests/public.rs
```diff
@@ -0,0 +1,101 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+extern crate mp4parse as mp4;
+
+use std::io::{Cursor, Read};
+use std::fs::File;
+
+// Took from https://github.com/GuillaumeGomez/audio-video-metadata/blob/9dff40f565af71d5502e03a2e78ae63df95cfd40/src/metadata.rs#L53
+//
+// Not important if it fails or not. If it compiles, it means it's all good.
```
rillian: Do these comments still apply?

tests/public.rs
```diff
@@ -0,0 +1,101 @@
...
+                        }
+                    }, "ES");
+                    assert!(a.samplesize > 0);
+                    assert!(a.samplerate > 0);
+                }
+                Some(mp4::SampleEntry::Unknown) | None => {}
+            }
+        }
+    } else {
+        panic!("read_mp4 function failed")
```
rillian: Same here, but more so. `if let Ok(())` is cute but the `else` clause is far removed from its conditional. Early returns are better.

rillian:
Thanks for adding a more comprehensive test! Please address the comments and update the PR.

tests/public.rs
GuillaumeGomez: Ah damn, no. I forgot to remove it.

tests/public.rs
GuillaumeGomez: For me, the readability is at its best here. But if you prefer another match, it's fine by me.

tests/public.rs
GuillaumeGomez: Ok!

GuillaumeGomez:
Updated.

Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.4.1"
 authors = [
   "Ralph Giles <giles@mozilla.com>",
   "Matthew Gregan <kinetik@flim.org>",
+  "Guillaume Gomez <guillaume1.gomez@gmail.com>"
```
rillian: If you don't mind I think it makes more sense to keep the author's list to just the two primary contacts. Your contributions will be clear in the commit history.

rillian:
Thanks, that's better. One more comment to address and then I'm happy to merge.

Cargo.toml
GuillaumeGomez: Fine by me! I remove it.

GuillaumeGomez:
Updated.

GuillaumeGomez:
Updated.

GuillaumeGomez:
When do you think this version will be published?

rillian:
I was hoping to get the other PRs merged first, but I can publish what we have now if that's helpful.

rillian:
I don't know if making more symbols public counts as a minor or patch update under semver, but we're on 0.x so it doesn't matter! I've published https://crates.io/crates/mp4parse/0.4.1 with your change. Hope that's enough to get your code working.

GuillaumeGomez:
Thanks!

PR #21, opened 2016-08-30
Split project into two crates mp4parse and mp4parse_capi (fixes #18)

### LINKED ISSUE #18: Put the capi module behind a feature flag
Servo doesn't need the capi module, and this brings rusty-cheddar and syntex to our already quite big dependency graph. Could that thing be put behind a feature flag?

nox:


nox:
Rebased and ready to be merged. :)

mp4parse_capi/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+# Cargo includes random files from the working directory
+# by default! Avoid bloating the package with test files.
+exclude = [
+  "*.mp4",
+]
+
+build = "build.rs"
+
+[dependencies]
+"mp4parse" = {path = "../mp4parse"}
```
rillian: So the idea would be to rewrite this to a version number every time I want to publish a new version of the crates?

At least `cargo publish` complains, so it's not possible to forget.

nox: What do you mean it complains?

rillian:
```
$ cargo publish
    Updating registry `https://github.com/rust-lang/crates.io-index`
error: all path dependencies must have a version specified when publishing.
dependency `mp4parse` does not specify a version
```

I filed https://github.com/rust-lang/cargo/pull/3060 to add the same check to `cargo package`.

nox: Oh right! https://github.com/serde-rs/serde/blob/master/serde_codegen/Cargo.toml#L36

This is not a bug and it indeed needs a version number.

PR #24, opened 2016-09-08
Use checked arithmetic in time_to_us.

rillian:
Since movie and track durations and media times can be 64-bit
integers, converting them to microseconds can overflow. Fortunately
our runtime error checking caught an instance of this.

Change the return type of the conversion routines to an Option and
use checked artimetic. In get_track_info default to 0 for overflowing
media_time and return an error for overflowing duration. This is
sufficient to accept the triggering content.

Rename time_to_ms to time_to_us to clarify that they return times
in microseconds.

See also https://bugzilla.mozilla.org/show_bug.cgi?id=1301065

### DIFF
--- a/mp4parse_capi/src/lib.rs
+++ b/mp4parse_capi/src/lib.rs
@@ -278,15 +278,36 @@ pub unsafe extern fn mp4parse_get_track_count(parser: *const mp4parse_parser, co
     MP4PARSE_OK
 }
 
-fn media_time_to_ms(time: MediaScaledTime, scale: MediaTimeScale) -> u64 {
-    assert!(scale.0 != 0);
-    time.0 * 1000000 / scale.0
+/// Calculate numerator * scale / denominator, if possible.
+///
+/// Applying the associativity of integer arithmetic, we divide first
+/// and add the remainder after multiplying each term separately
+/// to preserve precision while leaving more headroom. That is,
+/// (n * s) / d is split into floor(n / d) * s + (n % d) * s / d.
+///
+/// Return None on overflow or if the denominator is zero.
+fn rational_scale(numerator: u64, denominator: u64, scale: u64) -> Option<u64> {
+    if denominator == 0 {
+        return None;
+    }
+    let integer = numerator / denominator;
+    let remainder = numerator % denominator;
+    match integer.checked_mul(scale) {
+        Some(integer) => remainder.checked_mul(scale)
+            .and_then(|remainder| (remainder/denominator).checked_add(integer)),
+        None => None,
+    }
 }
 
-fn track_time_to_ms(time: TrackScaledTime, scale: TrackTimeScale) -> u64 {
+fn media_time_to_us(time: MediaScaledTime, scale: MediaTimeScale) -> Option<u64> {
+    let microseconds_per_second = 1000000;
+    rational_scale(time.0, scale.0, microseconds_per_second)
+}
+
+fn track_time_to_us(time: TrackScaledTime, scale: TrackTimeScale) -> Option<u64> {
     assert!(time.1 == scale.1);
-    assert!(scale.0 != 0);
-    time.0 * 1000000 / scale.0
+    let microseconds_per_second = 1000000;
+    rational_scale(time.0, scale.0, microseconds_per_second)
 }
 
 /// Fill the supplied `mp4parse_track_info` with metadata for `track`.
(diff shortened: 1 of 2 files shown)


rillian:
r? @kinetiknz 

This seems sufficient to handle the periscope stream. I don't know if falling back to zero would be better for the duration as well?

mp4parse_capi/src/lib.rs
```diff
@@ -278,15 +278,15 @@ pub unsafe extern fn mp4parse_get_track_count(parser: *const mp4parse_parser, co
...
-    time.0 * 1000000 / scale.0
+    time.0.checked_mul(1000000).and_then(|t| t.checked_div(scale.0))
 }
 
-fn track_time_to_ms(time: TrackScaledTime, scale: TrackTimeScale) -> u64 {
+fn track_time_to_us(time: TrackScaledTime, scale: TrackTimeScale) -> Option<u64> {
     assert!(time.1 == scale.1);
     assert!(scale.0 != 0);
-    time.0 * 1000000 / scale.0
+    time.0.checked_mul(1000000).and_then(|t| t.checked_div(scale.0))
```
kinetiknz: I think we can do better than this if we take a similar approach to Chromium's TimeDeltaFromRational, e.g. https://crbug.com/345888 and the related change to track_run_iterator.cc.

I'm r-ed the libstagefright changes for the same reason.

mp4parse_capi/src/lib.rs
```diff
@@ -334,12 +334,17 @@ pub unsafe extern fn mp4parse_get_track_info(parser: *mut mp4parse_parser, track
                                      context.timescale,
                                      track.duration) {
         info.media_time = track.media_time.map_or(0, |media_time| {
-            track_time_to_ms(media_time, track_timescale) as i64
+            track_time_to_us(media_time, track_timescale)
+                .unwrap_or(0) as i64
```
kinetiknz: You can't use unwrap_or here, this failure needs to turn into an MP4PARSE_ERROR return.

mp4parse_capi/src/lib.rs
```diff
@@ -334,12 +334,17 @@ pub unsafe extern fn mp4parse_get_track_info(parser: *mut mp4parse_parser, track
...
         }) - track.empty_duration.map_or(0, |empty_duration| {
-            media_time_to_ms(empty_duration, context_timescale) as i64
+            media_time_to_us(empty_duration, context_timescale)
+                .unwrap_or(0) as i64
```
kinetiknz: Ditto.

kinetiknz:
r- on the basis of the comments above.

mp4parse_capi/src/lib.rs
```diff
@@ -278,15 +278,15 @@ pub unsafe extern fn mp4parse_get_track_count(parser: *const mp4parse_parser, co
...
+fn media_time_to_us(time: MediaScaledTime, scale: MediaTimeScale) -> Option<u64> {
     assert!(scale.0 != 0);
-    time.0 * 1000000 / scale.0
+    time.0.checked_mul(1000000).and_then(|t| t.checked_div(scale.0))
 }
 
-fn track_time_to_ms(time: TrackScaledTime, scale: TrackTimeScale) -> u64 {
+fn track_time_to_us(time: TrackScaledTime, scale: TrackTimeScale) -> Option<u64> {
     assert!(time.1 == scale.1);
     assert!(scale.0 != 0);
```
kinetiknz: This assert becomes unnecessary with the used of checked_div.

mp4parse_capi/src/lib.rs
rillian: I thought we couldn't reject the file. It's not invalid, just inconvenient. Returning an error here would fix the panic, but at the expense of not playing a file we need to be able to play.

rillian:
I'll see if I can work up something based on the chrome class. I'm at a conference, so I don't know how far I'll get before next week.

mp4parse_capi/src/lib.rs
kinetiknz: If track_time_to_us does the right thing, the file isn't rejected.

rillian: Ah, I see what you mean. We could still end up rejecting technically valid files though, even if we pass this particular one.

rillian:
r=kinetik via https://bugzilla.mozilla.org/show_bug.cgi?id=1301065#a486109_274575

PR #25, opened 2016-09-14
Use corrected mimetype for vp9, vp8 and opus.

alfredoyang:


rillian:
This is setting the track mimetype, for which I think the bare codec is appropriate. What do you think about prepending the container either in the capi wrapper or in the C++ caller in gecko?

You should do avc as well to be consistent.

kinetiknz:
We should probably get rid of the internal MIME type entirely and replace it with a codec enum.  We already did that at the Rust/C interface to avoid dealing with copying Rust strings out, so the existence internally is a historical artifact copied from the libstagefright code (thus the reference in the comments to matching libstagefright's internal MIME strings).

rillian:
Good point. Thanks, @kinetiknz.

alfredoyang:
There is an enum mp4parse_codec in mp4parse_capi and codec_specific is used to retrieve the codec type, not mime_type. I don't see other reason to keep it. Maybe we should remove it?

rillian:
Hmm, maybe!

Servo is using the pure rust API, so we do need to make some effort to keep it clean. `Track::mime_type` is public, but it looks like they're not using it either and instead convert to their own scheme just as we do in mp4parse_capi. Seems like unnecessary duplication, but maybe there's too much mismatch between the requirements of different clients?

 So we could move the capi codec enum into the rust api if people would use that, or remove it until we get further feedback. @GuillaumeGomez do you have a preference?

GuillaumeGomez:
@rillian: I'll use them in a more or less close future. For now I don't but please have/keep it on rust side.

rillian:
Ok, sounds like converting the current strings on the rust side to an enum is the best approach.

mp4parse/src/lib.rs
```diff
@@ -291,6 +291,22 @@ impl Default for TrackType {
...
+#[derive(Debug)]
+pub enum CodecType {
+    Unknown,
+    AAC,
+    OPUS,
+    H264,
+    VP9,
+    VP8,
+    EncyptedVideo,
+    EncyptedAudio,
```
kinetiknz: Typo: Encrypted, x2

mp4parse/src/lib.rs
```diff
@@ -291,6 +291,22 @@ impl Default for TrackType {
     fn default() -> Self { TrackType::Unknown }
 }
 
+#[derive(Debug)]
+pub enum CodecType {
+    Unknown,
+    AAC,
+    OPUS,
```
kinetiknz: enums don't need to be all caps, and Opus isn't an acronym, so make it "Opus"

mp4parse/src/lib.rs
```diff
@@ -1102,11 +1118,11 @@ fn read_hdlr<T: Read>(src: &mut BMFFBox<T>) -> Result<HandlerBox> {
...
-        BoxType::AVCSampleEntry | BoxType::AVC3SampleEntry => String::from("video/avc"),
-        BoxType::VP8SampleEntry => String::from("video/vp8"),
-        BoxType::VP9SampleEntry => String::from("video/vp9"),
-        BoxType::ProtectedVisualSampleEntry => String::from("video/crypto"),
+    track.codec_type = match name {
+        BoxType::AVCSampleEntry | BoxType::AVC3SampleEntry => CodecType::H264,
+        BoxType::VP8SampleEntry => CodecType::VP8,
+        BoxType::VP9SampleEntry => CodecType::VP9,
+        BoxType::ProtectedVisualSampleEntry => CodecType::EncyptedVideo,
         _ => return Err(Error::Unsupported("unhandled video sample entry type")),
```
kinetiknz: I think you can set codec_type to Unknown here now rather than returning an error

mp4parse/src/lib.rs
```diff
@@ -1176,12 +1192,11 @@ fn read_video_desc<T: Read>(src: &mut BMFFBox<T>, track: &mut Track) -> Result<S
...
+    track.codec_type = match name {
         // TODO(kinetik): stagefright inspects ESDS to detect MP3 (audio/mpeg).
-        BoxType::MP4AudioSampleEntry => String::from("audio/mp4a-latm"),
-        // TODO(kinetik): stagefright doesn't have a MIME mapping for this, revisit.
-        BoxType::OpusSampleEntry => String::from("audio/opus"),
-        BoxType::ProtectedAudioSampleEntry => String::from("audio/crypto"),
+        BoxType::MP4AudioSampleEntry => CodecType::AAC,
+        BoxType::OpusSampleEntry => CodecType::OPUS,
+        BoxType::ProtectedAudioSampleEntry => CodecType::EncyptedAudio,
         _ => return Err(Error::Unsupported("unhandled audio sample entry type")),
```
kinetiknz: I think you can set codec_type to Unknown here now rather than returning an error

mp4parse/src/lib.rs

mp4parse/src/lib.rs

mp4parse/src/lib.rs

kinetiknz:
Thanks!

PR #26, opened 2016-09-19
Don't use track_index as index of track vec in mp4parser_capi.

alfredoyang:
track_index is 0 based according to track type.

Add 'ref' in mp4parse_is_fragmented().

Pass mediasource mochitest dom/media/mediasource/test/test_BufferedSeek_mp4.html.

### DIFF
--- a/mp4parse/src/boxes.rs
+++ b/mp4parse/src/boxes.rs
@@ -55,4 +55,5 @@ box_database!(
     ProtectedVisualSampleEntry 0x656e6376, // "encv" - Need to check official name in spec.
     ProtectedAudioSampleEntry  0x656e6361, // "enca" - Need to check official name in spec.
     MovieExtendsBox            0x6d766578, // "mvex"
+    UserDataBox                0x75647461, // "udta"
 );
--- a/mp4parse_capi/src/lib.rs
+++ b/mp4parse_capi/src/lib.rs
@@ -52,6 +52,7 @@ use mp4parse::MediaScaledTime;
 use mp4parse::TrackTimeScale;
 use mp4parse::TrackScaledTime;
 use mp4parse::serialize_opus_header;
+use mp4parse::Track;
 
 // rusty-cheddar's C enum generation doesn't namespace enum members by
 // prefixing them, so we're forced to do it in our member names until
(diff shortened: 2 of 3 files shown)


mp4parse/src/boxes.rs
```diff
@@ -55,4 +55,5 @@ box_database!(
     ProtectedVisualSampleEntry 0x656e6376, // "encv" - Need to check official name in spec.
     ProtectedAudioSampleEntry  0x656e6361, // "enca" - Need to check official name in spec.
     MovieExtendsBox            0x6d766578, // "mvex"
+    UserDataBox                0x75647461, // "udta"
```
rillian: This would be better as a separate commit. It seems unrelated to the rest of the patch.

rillian:
I assume this will obsolete @kinetiknz's similar change in https://bugzilla.mozilla.org/show_bug.cgi?id=1301065 once is lands in gecko?

kinetiknz:
I'd rather not take this change and would prefer to keep the API as it is, since it makes more sense to me the way it was designed.  The fact that MP4Metadata wants to use a different method to index tracks can be dealt with in MP4Metadata per the patch in bug 1303247.

rillian:
I'm convinced by @kinetiknz's argument that we should keep the APIs separate. Feel free to open a separate PR for the user data change though.

mp4parse_capi/src/lib.rs
```diff
@@ -393,16 +394,15 @@ pub unsafe extern fn mp4parse_get_track_audio_info(parser: *mut mp4parse_parser,
...
-        TrackType::Audio => {}
-        _ => return MP4PARSE_ERROR_INVALID,
-    };
+    // track_index here is the audio track number counting from 0. So we collect all audio
+    // tracks and get the track_index-th track.
+    let iter = (&context.tracks).iter();
+    match iter.filter(|track| track.track_type == TrackType::Audio).nth(track_index as usize) {
+        Some(ref found) => track = found,
+        None => return MP4PARSE_ERROR_BADARG,
+    }
```
rillian: I thought this was very nice, btw.

PR #27, opened 2016-09-22
Parse mvex to retrieve fragment_duration.

alfredoyang:
Parse fragment_duration for https://bugzilla.mozilla.org/show_bug.cgi?id=1304254.

Test will be added in another PR later.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -265,13 +265,18 @@ pub struct OpusSpecificBox {
     channel_mapping_table: Option<ChannelMappingTable>,
 }
 
+#[derive(Debug)]
+pub struct MovieExtendsBox {
+    pub fragment_duration: Option<MediaScaledTime>,
+}
+
 /// Internal data structures.
 #[derive(Debug, Default)]
 pub struct MediaContext {
     pub timescale: Option<MediaTimeScale>,
-    pub has_mvex: bool,
     /// Tracks found in the file.
     pub tracks: Vec<Track>,
+    pub mvex: Option<MovieExtendsBox>,
 }
 
 impl MediaContext {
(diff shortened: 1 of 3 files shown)


mp4parse/src/lib.rs
```diff
@@ -565,6 +571,32 @@ fn read_moov<T: Read>(f: &mut BMFFBox<T>, context: &mut MediaContext) -> Result<
     Ok(())
 }
 
+fn read_mvex<T: Read>(src: &mut BMFFBox<T>) -> Result<MovieExtendsBox> {
+    let mut iter = src.box_iter();
+    let mut duration: Option<u64> = None;
```
rillian: The compiler should be able to infer this type from the signature of `read_mehd` so you only need this if you especially want to document it locally. Up to you.

mp4parse/src/lib.rs
```diff
@@ -565,6 +571,32 @@ fn read_moov<T: Read>(f: &mut BMFFBox<T>, context: &mut MediaContext) -> Result<
...
+            },
+            _ => try!(skip_box_content(&mut b)),
+        }
+    }
+    Ok(MovieExtendsBox {
+        fragment_duration: duration.map_or(None, |x| { Some(MediaScaledTime(x)) }),
+    })
+}
+
+fn read_mehd<T: Read>(src: &mut BMFFBox<T>) -> Result<Option<u64>> {
```
rillian: `mehd` always has a fragment duration, so you don't need to return an Option here. A simpler `Result<u64>` or `Result<MediaScaledTime>` would work. This complicates the way you update duration, but I think it's better to have a temporary there and keep the semantics of the function signature correct.

e.g.
```rust
let mut fragment_duration = None;
...
            let duration = try!(read_mehd(&mut b));
            fragment_duration = Some(duration);
...
Ok(MovieExtendsBox {
    fragment_duration: fragment_duration,
})
```

mp4parse_capi/src/lib.rs
```diff
@@ -503,8 +536,9 @@ pub unsafe extern fn mp4parse_is_fragmented(parser: *mut mp4parse_parser, track_
     let tracks = &context.tracks;
     (*fragmented) = false as u8;
 
-    if !context.has_mvex {
-        return MP4PARSE_OK;
+    match context.mvex {
+        Some(_) => {},
+        None => return MP4PARSE_OK,
```

mp4parse/src/lib.rs

mp4parse/src/lib.rs

alfredoyang:
Should I update parser to gecko in the same bug? Because there is another gecko patch depending on this in the same bug.

rillian:
If you need these changes in gecko we can bump the in-tree version on the bugzilla side. You can do it, or I can, I've had lots of practice.

PR #31, opened 2016-09-30
Skip padding data

alfredoyang:
This is from https://bugzilla.mozilla.org/show_bug.cgi?id=1306236.

Some contents may have extra padding data in the boxes of stbl. For example, stsc, stsz, stsd...etc.
This patch skips these paddings so parser keeps parsing the file.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -469,6 +469,17 @@ fn skip_box_content<T: Read>(src: &mut BMFFBox<T>) -> Result<()> {
     skip(src, to_skip)
 }
 
+/// Skip over the remain data of a box.
+fn skip_box_remain<T: Read>(src: &mut BMFFBox<T>) -> Result<()> {
+    let remain = {
+        let header = src.get_header();
+        let len = src.bytes_left();
+        log!("remain {} (skipped) in {:?}", len, header);
+        len
+    };
+    skip(src, remain)
+}
+
 macro_rules! check_parser_state {
     ( $src:expr ) => {
         if $src.limit() > 0 {
@@ -921,6 +932,9 @@ fn read_stco<T: Read>(src: &mut BMFFBox<T>) -> Result<ChunkOffsetBox> {
         offsets.push(try!(be_u32(src)) as u64);
     }
 
+    // Padding could be added in some contents.
+    try!(skip_box_remain(src));
+
     Ok(ChunkOffsetBox {
         offsets: offsets,
     })
@@ -935,6 +949,9 @@ fn read_co64<T: Read>(src: &mut BMFFBox<T>) -> Result<ChunkOffsetBox> {
         offsets.push(try!(be_u64(src)));
     }
 
+    // Padding could be added in some contents.
+    try!(skip_box_remain(src));
+
     Ok(ChunkOffsetBox {
         offsets: offsets,
     })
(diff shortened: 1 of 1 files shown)


mp4parse/src/lib.rs
```diff
@@ -1311,38 +1340,43 @@ fn read_stsd<T: Read>(src: &mut BMFFBox<T>, track: &mut Track) -> Result<SampleD
...
-        let description = match description {
-            Ok(desc) => desc,
-            Err(Error::Unsupported(_)) => {
-                // read_{audio,video}_desc may have returned Unsupported
-                // after partially reading the box content, so we can't
-                // simply use skip_box_content here.
-                let to_skip = b.bytes_left();
-                try!(skip(&mut b, to_skip));
-                SampleEntry::Unknown
+    {
```
rillian: I assume the extra indent so `iter` releases it's reference to src for the subsequent skip? I don't see an alternate solution besides moving this code into a `read_description` helper function. That might not be a bad idea for readability anyway.

mp4parse/src/lib.rs
```diff
@@ -1311,38 +1340,43 @@ fn read_stsd<T: Read>(src: &mut BMFFBox<T>, track: &mut Track) -> Result<SampleD
...
-        }
-        descriptions.push(description);
-        check_parser_state!(b.content);
-        if descriptions.len() == description_count as usize {
-            break;
         }
     }
 
+    // Padding could be added in some contents.
+    try!(skip_box_remain(src));
```
rillian: This will skip additional entries where `description_count` is short. That's probably fine.

rillian:
Test failure seems to be an infrastructure problem. I'll go ahead and merge; please file new PRs for followup issues.

mp4parse/src/lib.rs
alfredoyang: Yes, it releases the borrow for the following skip.

alfredoyang:
I'll add test at another PR.

I can't find any statement about the extra padding in 14496-12; however, there are 'free' and 'skip' box defined in spec. I think it should be a special case.

rillian:
Ok, thanks.

alfredoyang:
I check the bitstream again. These paddings are 'skip' boxes actually.
The layout is:

stsd box
  ...
  ...
  skip box
  ...
  skip box
  ...
stsc box
  ...
  ...
  skip box
stsz box
  ...
  skip box

I don't think it is correct behaviour because 'stsd' is not a container box. My dump tool MP4Box can't parse this kind of content correctly neither.

There are 2 ways in my mind currently.
1. Add a function to jump over 'skip' box at the end of read_stsd, read_stsc...etc. Just like this pull.
2. Jump over 'skip' box in check_parser_state(). So all the 'skip' boxes will be omitted in read_stbl.

Any idea?

rillian:
Hmm. I think this is just broken software. These (`stsd`, `stsc`, `stsz`) are all variable-length boxes. Maybe it leaves some padding while writing the header and then fills in skip boxes later?

Do we need to do more than we did with this PR? Skipping to the end of these boxes, regardless of the contents seems safer than trying to interpret them, when we want compatibility with what stagefright accepts.

alfredoyang:
Ok, we don't need to do anything for these contents. I'll submit another PR for test later.

PR #33, opened 2016-10-03
Update flac parser for the 0.0.2 draft specification.

rillian:
FLACSpecificBox is now a fullbox.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -1107,9 +1107,12 @@ fn read_flac_metadata<T: Read>(src: &mut BMFFBox<T>) -> Result<FLACMetadataBlock
 
 /// Parse `FLACSpecificBox`.
 fn read_dfla<T: Read>(src: &mut BMFFBox<T>) -> Result<FLACSpecificBox> {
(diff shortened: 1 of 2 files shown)


mp4parse/src/lib.rs
```diff
@@ -1107,10 +1107,13 @@ fn read_flac_metadata<T: Read>(src: &mut BMFFBox<T>) -> Result<FLACMetadataBlock
 
 /// Parse `FLACSpecificBox`.
 fn read_dfla<T: Read>(src: &mut BMFFBox<T>) -> Result<FLACSpecificBox> {
-    let version = try!(src.read_u8());
+    let (version, flags) = try!(read_fullbox_extra(src));
     if version != 0 {
         return Err(Error::Unsupported("unknown dfLa (FLAC) version"));
     }
+    if flags != 0x000000 {
```
kinetiknz: Might as well just be `flags != 0`.

rillian: The spec uses this to make comparing 24-bit values easier. But we never expect to use the flags. I'll change it.

PR #36, opened 2016-10-07
Add test for skip padding data.

alfredoyang:


mp4parse/src/tests.rs
```diff
@@ -757,3 +757,68 @@ fn invalid_pascal_string() {
...
+            _ => (),
+        }
+    }
+}
+
+#[test]
+fn skip_padding_in_stsd() {
+    // Padding data could be added in the end of stsd boxes. Parser needs to skip
+    // them instead of returning error.
+    let make_avc = make_box(BoxSize::Auto, b"avc1", |s| {
```
rillian: Better to call this `avc`. We've been using the verb 'make' for functions producing a box, rather than the result.

alfredoyang: Fixed.

Issue #30, opened 2016-09-30
Implement `esds` parsing


kinetiknz:
The `media.rust.test_mode` pref in Gecko found another mismatch between mp4parse and libstagefright.  From a Facebook live stream, a `moov` with a video and audio track results in mp4parse returning a samplerate of 44100 and libstagefright returning a samplerate of 22050.

mp4parse is parsing the samplerate directly from the `mp4a`.  libstagefright prefers the value from the `esds`.

alfredoyang:
hmm... it looks like not just audio sample rate but also channel count needs to be retrieved from esds in Facebook video. I'll have another PR for it.

jyavenard:
Actually, 44.1kHz would be the value to use. The ESDS value returned is of the AAC-LC stream. 
Facebook uses HE-AAC v2. 

Sampling rate would be twice as much as the LC stream. 

The stage fright returns the sampling rate of the LC stream is an absolute pain in the arse as it requires to decode the first sample to get the real sampling rate. 

Ideally, for AAC, the AudioSpecificConfig should be decoded and detect the presence of SBR. 

The sampling rate value found in the ESDS Is useless

rillian:
@jyavenard So we should drop the ESDS value from both the rust and stagefright parsers?

kinetiknz:
At this stage we're trying to match stagefright's behaviour (as observable via the API we're using in the MP4Metadata class) so that mp4parse-rust works as a drop-in replacement.  Making different choices for which metadata to expose and from which source is possible and may make sense in the future, but it's outside the scope of what we're trying to do here.

kinetiknz:
Closing as obsolete.  PR #38 added ESDS parsing and PR #39 addressed the functionality this issue was about (matching behaviour with libstagefright).

PR #38, opened 2016-10-13
Parse ESDS.

alfredoyang:
Structure of ESDS:

  ESDS box
   |
   |- ES descriptor
   |   |- 3 bytes extension (optional)
   |   |- 1 byte length
   |   |- 2 bytes ES ID
   |   |- 1 byte stream priority and flags
   |   |- stream dependency (optional, depends on flags)
   |   |   |- 2 bytes
   |   |- url (optional, depends on flags)
   |   |   |- 1 byte length
   |   |   |- x bytes, x is the length
   |   |
   |   |- Decoder Config descriptior
   |   |- 3 bytes extension (optional)
   |   |- 1 byte length
   |   |- 1 byte object profile indicator

We need to get object profile indicator for the codec type.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -11,6 +11,7 @@ extern crate afl;
 extern crate byteorder;
 use byteorder::ReadBytesExt;
 use std::io::{Read, Take};
+use std::io::Cursor;
 use std::cmp;
 
 mod boxes;
@@ -22,6 +23,10 @@ mod tests;
 
 // Arbitrary buffer size limit used for raw read_bufs on a box.
 const BUF_SIZE_LIMIT: u64 = 1024 * 1024;
+
+// Tags for elementary stream description
+const ESDESCR_TAG: u8           = 0x03;
+const DECODER_CONFIG_TAG: u8    = 0x04;
 
 static DEBUG_MODE: std::sync::atomic::AtomicBool = std::sync::atomic::ATOMIC_BOOL_INIT;
 
@@ -203,10 +208,16 @@ pub enum SampleEntry {
     Unknown,
 }
 
+#[derive(Debug, Clone)]
+pub struct EsDescriptor {
+    pub audio_codec: CodecType,
+    pub codec_specific_config: Vec<u8>,
+}
+
 #[allow(non_camel_case_types)]
 #[derive(Debug, Clone)]
 pub enum AudioCodecSpecific {
-    ES_Descriptor(Vec<u8>),
+    EsDescriptor(EsDescriptor),
     FLACSpecificBox(FLACSpecificBox),
     OpusSpecificBox(OpusSpecificBox),
 }
@@ -310,9 +321,10 @@ impl Default for TrackType {
     fn default() -> Self { TrackType::Unknown }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub enum CodecType {
     Unknown,
+    MP3,
     AAC,
     FLAC,
     Opus,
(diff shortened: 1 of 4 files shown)


kinetiknz:
We also want to get the "real" sample rate out of the ESDS to fix issue #30, but that can be done separately to this PR.

alfredoyang:
Ok, I'll address issue #30 later.  Do we also need to parse ESDS in VideoSampleEntry?

kinetiknz:
>  Do we also need to parse ESDS in VideoSampleEntry?

Probably not at this stage.  libstagefright only parses it to extract accurate audio metadata; for video it's just exporting the entire ESDS as a binary blob without inspecting it.

mp4parse/src/lib.rs
```diff
@@ -1105,16 +1117,77 @@ fn read_flac_metadata<T: Read>(src: &mut BMFFBox<T>) -> Result<FLACMetadataBlock
...
 
     let esds_size = src.head.size - src.head.offset - 4;
     if esds_size > BUF_SIZE_LIMIT {
         return Err(Error::InvalidData("esds box exceeds BUF_SIZE_LIMIT"));
     }
-    let esds = try!(read_buf(&mut src.content, esds_size as usize));
+    let esds_array = try!(read_buf(src, esds_size as usize));
+
+    // clone a esds cursor for parsing.
+    let esds = &mut Cursor::new(esds_array.clone());
```
kinetiknz: Rather than cloning the data, share it by creating the Cursor with a ref.  You just need to arrange for the Cursor to goe out of scope before you try to move esds_array into EsDescriptor.

mp4parse/src/lib.rs
```diff
@@ -1398,7 +1471,8 @@ fn read_audio_sample_entry<T: Read>(src: &mut BMFFBox<T>, track: &mut Track) ->
                 }
 
                 let esds = try!(read_esds(&mut b));
-                codec_specific = Some(esds);
+                track.codec_type = esds.audio_codec;
```
kinetiknz: For consistency, set track.codec_type and codec_specific in the same order in each of the branches - right now there's a mix of both variants.

mp4parse/src/lib.rs
```diff
@@ -23,6 +24,10 @@ mod tests;
 // Arbitrary buffer size limit used for raw read_bufs on a box.
 const BUF_SIZE_LIMIT: u64 = 1024 * 1024;
 
+// Tags for elementary stream description
+const ESDESCR_TAG: u8           = 0x03;
+const DECODER_CONFIG_TAG: u8    = 0x04;
```
kinetiknz: These are only used inside `read_esds`, so move them into the scope of the function.

mp4parse/src/lib.rs
```diff
@@ -203,10 +208,16 @@ pub enum SampleEntry {
...
+pub struct EsDescriptor {
+    pub audio_codec: CodecType,
+    pub codec_specific_config: Vec<u8>,
+}
+
 #[allow(non_camel_case_types)]
 #[derive(Debug, Clone)]
 pub enum AudioCodecSpecific {
-    ES_Descriptor(Vec<u8>),
+    EsDescriptor(EsDescriptor),
```
kinetiknz: Keep the name ES_Descriptor to match the spec reference.

mp4parse/src/lib.rs
```diff
@@ -1105,16 +1117,77 @@ fn read_flac_metadata<T: Read>(src: &mut BMFFBox<T>) -> Result<FLACMetadataBlock
...
+    let esds_tag = try!(esds.read_u8());
+
+    if esds_tag != ESDESCR_TAG {
+        return Err(Error::Unsupported("fail to parse ES descriptor"));
+    }
+
+    let esds_extend = try!(esds.read_u8());
+    // extension tag start from 0x80.
+    if esds_extend >= 0x80 {
+        // skip remains extension and length.
```
kinetiknz: remaining

mp4parse/src/lib.rs
```diff
@@ -1105,16 +1117,77 @@ fn read_flac_metadata<T: Read>(src: &mut BMFFBox<T>) -> Result<FLACMetadataBlock
...
+        try!(skip(esds, 5));
+    } else {
+        try!(skip(esds, 2));
+    }
+
+    let esds_flags = try!(esds.read_u8());
+
+    // Stream dependnecy flag, first bit from left most.
```
kinetiknz: dependency

mp4parse/src/lib.rs
```diff
@@ -1105,16 +1117,77 @@ fn read_flac_metadata<T: Read>(src: &mut BMFFBox<T>) -> Result<FLACMetadataBlock
...
+    if esds_flags & 0x80 > 0 {
+        // Skip uninteresting fields.
+        try!(skip(esds, 2));
+    }
 
-    Ok(AudioCodecSpecific::ES_Descriptor(esds))
+    // Url flag, second bit from left most.
+    if esds_flags & 0x40 > 0 {
+        // Skip uninteresting fields.
+        let skip_es_len = try!(esds.read_u8()) + 2;
```
kinetiknz: It'd be safer to make skip_es_len a usize and cast the result of the read before adding to avoid potential overflows.

mp4parse/src/lib.rs

mp4parse/src/lib.rs

mp4parse/src/lib.rs

mp4parse/src/lib.rs

mp4parse/src/lib.rs

mp4parse/src/lib.rs

mp4parse/src/lib.rs

kinetiknz:
Great, thanks!

PR #39, opened 2016-10-20
Parse codec specific descriptor to retrieve audio sample rate

alfredoyang:


alfredoyang:
This is for https://github.com/mozilla/mp4parse-rust/issues/30.

mp4parse/src/lib.rs
```diff
@@ -1128,24 +1137,30 @@ fn read_esds<T: Read>(src: &mut BMFFBox<T>) -> Result<ES_Descriptor> {
...
-        let esds_tag = try!(esds.read_u8());
+        let mut next_tag = try!(esds.read_u8());
 
-        if esds_tag != ESDESCR_TAG {
+        if next_tag != ESDESCR_TAG {
             return Err(Error::Unsupported("fail to parse ES descriptor"));
         }
 
         let esds_extend = try!(esds.read_u8());
+        let mut esds_end: u64 = 0;
```
rillian: Rather than turning off the warning with `#[allow(unused_assignments)]`, you can avoid it with an `if` expression like a ternary in C. I'm not sure this is actually easier to read, but avoiding the `mut` is more natural in rust, I think.

```rust
let esds_extend = try!(esds.read_u8());
// extension tag starts from 0x80.
let esds_end = if esds_extend >= 0x80 {
    // skip remaining extension.
    try!(skip(esds, 2));
    esds.position() + try!(esds.read_u8()) as u64
} else {
    esds.position() + esds_extend as u64
};
try!(skip(esds, 2));
```

mp4parse/src/lib.rs
```diff
@@ -1163,19 +1178,45 @@ fn read_esds<T: Read>(src: &mut BMFFBox<T>) -> Result<ES_Descriptor> {
         }
 
         // find DecoderConfig descriptor (tag = DECODER_CONFIG_TAG)
-        let dcds = try!(esds.read_u8());
-        if dcds != DECODER_CONFIG_TAG {
-            return Err(Error::Unsupported("fail to parse decoder config descriptor"));
+        if esds_end > esds.position() {
+            next_tag = try!(esds.read_u8());
```
rillian: Since these uses of `next_tag` start with a fresh assignment within a block, please drop the `mut` keyword from the outer declaration and use a fresh `let` here and below. Rust will not warn about the shadowing, and using a local in each block reduces the amount of code which must be understood at once.

mp4parse_capi/src/lib.rs
```diff
@@ -444,6 +444,12 @@ pub unsafe extern fn mp4parse_get_track_audio_info(parser: *mut mp4parse_parser,
             }
             (*info).codec_specific_config.length = v.codec_specific_config.len() as u32;
             (*info).codec_specific_config.data = v.codec_specific_config.as_ptr();
+            match v.audio_sample_rate {
+                Some(rate) => {
+                    (*info).sample_rate = rate;
+                },
+                _ => {},
+            }
```
alfredoyang: Thanks for suggestion. Fixed.

mp4parse/src/lib.rs

PR #40, opened 2016-10-21
check channelcounts in CSD

alfredoyang:
Audio channel count in ESDS is different comparing to AudioSampleEntry for Facebook video. Stagefright uses audio channel count from ESDS.
So MP4 rust parser should be able to retrieve it.

https://bugzilla.mozilla.org/show_bug.cgi?id=1311929

### DIFF
--- a/mp4parse_capi/src/lib.rs
+++ b/mp4parse_capi/src/lib.rs
@@ -447,6 +447,9 @@ pub unsafe extern fn mp4parse_get_track_audio_info(parser: *mut mp4parse_parser,
             if let Some(rate) = v.audio_sample_rate {
                 (*info).sample_rate = rate;
             }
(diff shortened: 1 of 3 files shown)


mp4parse_capi/src/lib.rs
```diff
@@ -792,7 +795,7 @@ fn arg_validation_with_data() {
 
         let mut audio = Default::default();
         assert_eq!(MP4PARSE_OK, mp4parse_get_track_audio_info(parser, 1, &mut audio));
-        assert_eq!(audio.channels, 2);
+        assert_eq!(audio.channels, 1);
```
rillian: You need to make the same change in mp4parse_capi/examples/test.cc.

PR #42, opened 2016-10-25
Some minor updates to the flac parser

rillian:


mp4parse/src/lib.rs
```diff
@@ -257,16 +257,16 @@ pub struct VPxConfigBox {
 }
 
 #[derive(Debug, Clone)]
-struct FLACMetadataBlock {
+pub struct FLACMetadataBlock {
     block_type: u8,
```
kinetiknz: Don't you need the block_type to identify the block?

mp4parse/tests/public.rs
```diff
@@ -75,8 +75,9 @@ fn public_api() {
                         assert_eq!(esds.audio_sample_rate.unwrap(), 48000);
                         "ES"
                     }
-                    mp4::AudioCodecSpecific::FLACSpecificBox(_) => {
-                        // No public fields.
+                    mp4::AudioCodecSpecific::FLACSpecificBox(flac) => {
+                        assert!(flac.blocks.len() > 0);
+                        assert!(flac.blocks[0].data.len() > 0);
```
kinetiknz: The first block is required to be STREAMINFO, so check that it's == 34?

mp4parse_capi/src/lib.rs
```diff
@@ -448,8 +448,14 @@ pub unsafe extern fn mp4parse_get_track_audio_info(parser: *mut mp4parse_parser,
                 (*info).sample_rate = rate;
             }
         }
-        AudioCodecSpecific::FLACSpecificBox(_) => {
-            return MP4PARSE_ERROR_UNSUPPORTED;
+        AudioCodecSpecific::FLACSpecificBox(ref flac) => {
+            // Return the STREAMINFO metadata block in the codec_specific.
+            let streaminfo = &flac.blocks[0].data;
+            if streaminfo.len() > std::u32::MAX as usize {
```
kinetiknz: Check the block_type and the size, since it's a known fixed value (34).

mp4parse/src/lib.rs
rillian: Yes, except that STREAMINFO is always first. I didn't want to expose it because I thought an enum was better, but that's what semver bumps are for.

mp4parse_capi/src/lib.rs
rillian: Good ideas, thanks.

PR #43, opened 2016-11-08
Parse 'pssh' box for eme playback.

alfredoyang:


mp4parse/src/lib.rs
```diff
@@ -601,13 +614,58 @@ fn read_moov<T: Read>(f: &mut BMFFBox<T>, context: &mut MediaContext) -> Result<
...
+        let system_id = try!(read_buf(pssh, 16));
+
+        let mut kid: Vec<ByteData> = Vec::new();
+        if version > 0 {
+            let mut count = try!(be_i32(pssh));
+            while count > 0 {
+                let item = try!(read_buf(pssh, 16));
+                kid.push(item);
+                count -= 1;
+            }
```
rillian: I think `for _ in 0..count` would be better than a `while` loop here. It's a little awkward, but there's one less line of code, and `count` doesn't need to be `mut`.

mp4parse_capi/src/lib.rs
```diff
@@ -581,6 +603,40 @@ pub unsafe extern fn mp4parse_is_fragmented(parser: *mut mp4parse_parser, track_
...
+        match data_len.write_u32::<byteorder::NativeEndian>(pssh.box_content.len() as u32) {
+            Err(_) => {
+                return MP4PARSE_ERROR_IO;
+            },
+            _ => (),
+        }
+        // pssh data format in gecko:
+        //   system_id
+        //   pssh size (in native endian)
+        //   pssh box content (including header)
```
rillian: It might be better to put this description in the function-level doc-comment above, since it's something a caller would need to know, not just an internal implementation detail.

mp4parse/src/lib.rs
```diff
@@ -601,13 +614,58 @@ fn read_moov<T: Read>(f: &mut BMFFBox<T>, context: &mut MediaContext) -> Result<
...
+
+        let data_size = try!(be_i32(pssh)) as usize;
+        let data = try!(read_buf(pssh, data_size));
+
+        (system_id, kid, data)
+    };
+
+    let mut pssh_box = Vec::new();
+    try!(write_be_u32(&mut pssh_box, src.head.size as u32));
+    pssh_box.append(&mut vec![b'p', b's', b's', b'h']);
```
rillian: You can `pssh_box.extend(b"pssh");` which is a bit easier to read.

NB there's also `Vec::extend_from_slice()` but `extend` is equally fast in rust 1.14, so I don't think it's worth using the longer method name here. https://github.com/rust-lang/rust/pull/37094

alfredoyang:
Thanks for review. All comments addressed.

rillian:
Thanks!

PR #46, opened 2016-11-23
Separate read_esds() into smaller functions.

alfredoyang:
The old read_esds() is too long. It is hard to maintain. So I think it'd be better to separate it into small functions.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -1173,7 +1173,7 @@ fn read_flac_metadata<T: Read>(src: &mut BMFFBox<T>) -> Result<FLACMetadataBlock
     })
 }
 
-fn find_descriptor<'a>(data: &'a [u8], esds: &mut ES_Descriptor) -> Result<()> {
+fn find_descriptor(data: &[u8], esds: &mut ES_Descriptor) -> Result<()> {
     // Tags for elementary stream description
     const ESDESCR_TAG: u8          = 0x03;
     const DECODER_CONFIG_TAG: u8   = 0x04;
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -205,12 +205,12 @@ pub enum SampleEntry {
 }
 
 #[allow(non_camel_case_types)]
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub struct ES_Descriptor {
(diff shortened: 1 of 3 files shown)


mp4parse/src/tests.rs
```diff
@@ -827,7 +827,7 @@ fn skip_padding_in_stsd() {
 fn read_qt_wave_atom() {
     let esds = make_fullbox(BoxSize::Auto, b"esds", 0, |s| {
         s.B8(0x03)  // elementary stream descriptor tag
-         .B8(0x0b)  // esds length
+         .B8(0x12)  // esds length
```
rillian: Oops. Hex versus decimal length. How did this test pass before?

mp4parse/src/lib.rs
```diff
@@ -1173,129 +1173,141 @@ fn read_flac_metadata<T: Read>(src: &mut BMFFBox<T>) -> Result<FLACMetadataBlock
     })
 }
 
-fn read_esds<T: Read>(src: &mut BMFFBox<T>) -> Result<ES_Descriptor> {
+fn find_descriptor<'a>(data: &'a [u8], esds: &mut ES_Descriptor) -> Result<()> {
```
kinetiknz: The lifetime annotation here doesn't seem to be necessary.

rillian: I guess that's because the lifetime can safely propagate to the result.

I was more surprised by `let mut remains = data;` Why does that work?

kinetiknz: It's only the binding that's mutable; it allows `remains` to be reassigned in the loop.  `data` isn't becoming mutable; that'd be something like `let remains: &mut [u8] = data;`, which would be a compile error.

mp4parse/src/tests.rs
alfredoyang: Urg... the old method doesn't care the length very much.

Now the find_descriptor() checks the length to parse each descriptor; if the descriptor length is wrong, it will return failure.

mp4parse/src/lib.rs
alfredoyang: Yes, it doesn't need the lifetime here. I forgot to remove it during refactory.
Thanks for catching!

rillian: Ok, so at bit like `const *` vs `* const` in C. Thanks for clarifying.

PR #48, opened 2016-12-12
Parsing sinf box for enca/encv track

alfredoyang:
Parsing schi, tenc, sinf and frma boxes for https://bugzilla.mozilla.org/show_bug.cgi?id=1320026.


mp4parse/src/lib.rs
```diff
@@ -228,6 +228,7 @@ pub struct AudioSampleEntry {
     pub samplesize: u16,
     pub samplerate: u32,
     pub codec_specific: AudioCodecSpecific,
+    pub protect_info: Option<ProtectionSchemeInfoBox>,
```
rillian: I don't think this abbreviation is helpful. Please call this field `protection_info`.

mp4parse/src/lib.rs
```diff
@@ -242,6 +243,7 @@ pub struct VideoSampleEntry {
     pub width: u16,
     pub height: u16,
     pub codec_specific: VideoCodecSpecific,
+    pub protect_info: Option<ProtectionSchemeInfoBox>,
```
rillian: 'protection_info` here, likewise.

mp4parse/src/lib.rs
```diff
@@ -1537,14 +1563,6 @@ fn read_qt_wave_atom<T: Read>(src: &mut BMFFBox<T>) -> Result<ES_Descriptor> {
...
 fn read_audio_sample_entry<T: Read>(src: &mut BMFFBox<T>, track: &mut Track) -> Result<SampleEntry> {
     let name = src.get_header().name;
-    track.codec_type = match name {
-        // TODO(kinetik): stagefright inspects ESDS to detect MP3 (audio/mpeg).
-        BoxType::MP4AudioSampleEntry => CodecType::AAC,
-        BoxType::FLACSampleEntry => CodecType::FLAC,
-        BoxType::OpusSampleEntry => CodecType::Opus,
-        BoxType::ProtectedAudioSampleEntry => CodecType::EncryptedAudio,
-        _ => CodecType::Unknown,
-    };
```
rillian: You don't mention why you're removing this. I can see this might be dead code, in that `Unknown` is the default and `FLAC`, `Opus`, and `AAC` settings were added elsewhere with the esds parsing in #38. But please confirm it is dead code.

In general, cleanup like this is better in a separate commit so it's clear why the change was made.

mp4parse/src/lib.rs
```diff
@@ -1682,6 +1711,62 @@ fn read_stsd<T: Read>(src: &mut BMFFBox<T>, track: &mut Track) -> Result<SampleD
...
+fn read_sinf<T: Read>(src: &mut BMFFBox<T>) -> Result<ProtectionSchemeInfoBox> {
+    let mut sinf = ProtectionSchemeInfoBox::default();
+
+    let mut iter = src.box_iter();
+    while let Some(mut b) = iter.next_box()? {
+        match b.head.name {
+            BoxType::OriginalFormatBox => {
+                let frma = read_frma(&mut b)?;
+                sinf.code_name = frma;
+            },
```
rillian: I don't have the `cenc` spec, but 14496-12 says the `OriginalFormatBox` is required. Can we reject streams without one here? We don't seem to check later whether this is what we expect.

mp4parse/src/lib.rs
```diff
@@ -1682,6 +1711,62 @@ fn read_stsd<T: Read>(src: &mut BMFFBox<T>, track: &mut Track) -> Result<SampleD
...
+            BoxType::SchemeInformationBox => {
+                // We only need tenc box in schi box so far.
+                let schi_tenc = read_schi(&mut b)?;
+                sinf.tenc = schi_tenc;
+            },
```
rillian: Do we depend on the `tenc` data? If so we should probably also reject streams with a `sinf` without an `schi`.

mp4parse/src/lib.rs

mp4parse/src/lib.rs

mp4parse/src/lib.rs
alfredoyang: Yes, it is dead code. I'll separate that to another commit next time.

mp4parse/src/lib.rs
alfredoyang: Fixed

alfredoyang: fixed.

mp4parse/src/lib.rs
alfredoyang: fixed.

mp4parse/src/lib.rs
rillian: Ok.

PR #52, opened 2016-12-16
Initialize structures

alfredoyang:
mp4parser_capi acts as the bridge between rust and gecko. So the structures from gecko need to be initialized to default value in case of unused members.

### DIFF
--- a/mp4parse_capi/src/lib.rs
+++ b/mp4parse_capi/src/lib.rs
@@ -365,7 +365,7 @@ pub unsafe extern fn mp4parse_get_track_info(parser: *mut mp4parse_parser, track
         return MP4PARSE_ERROR_BADARG;
     }
 
-    // Set default value due to it is allocated in gecko.
+    // Initialize fields to default values to ensure all fields are always valid.
     *info = Default::default();
 
     let context = (*parser).context_mut();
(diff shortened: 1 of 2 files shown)


alfredoyang:
hmm... the assertion is on test.cc:180.
   177
   178 	  rv = mp4parse_get_track_info(parser, 3, &info);
   179 	  assert(rv == MP4PARSE_ERROR_BADARG);
-> 180 	  assert_zero(&info);
   181 	  rv = mp4parse_get_track_video_info(parser, 3, &video);
   182 	  assert(rv == MP4PARSE_ERROR_BADARG);


I'm not sure why it checks zero content here.

kinetiknz:
> I'm not sure why it checks zero content here.

It memset the struct to zero before passing it in, so it's (crudely) trying to test that the struct wasn't updated when MP4PARSE_ERROR_BADARG was returned.

You could argue that the caller shouldn't examine the struct at all after an error is returned since it may be in an indeterminate state (which is likely if the error happens part-way through the struct's initialization).  The test was using knowledge of the implementation to enforce something stricter, but maybe it doesn't make sense to do that and we should just remove those tests.

mp4parse_capi/src/lib.rs
```diff
@@ -354,6 +365,9 @@ pub unsafe extern fn mp4parse_get_track_info(parser: *mut mp4parse_parser, track
         return MP4PARSE_ERROR_BADARG;
     }
 
+    // Set default value due to it is allocated in gecko.
```
kinetiknz: Maybe make this "Initialize fields to default values to ensure all fields are always valid."?

alfredoyang:
Thanks for review!

PR #54, opened 2017-01-03
Support better error msg

alfredoyang:
These changes are for [1] to find bug easier.

1. Change 4CC number to characters in logs.

For example:

FileTypeBox { major_brand: "senv", minor_version: 0, compatible_brands: ["senv", "iso6", "dash"] }
MovieHeaderBox { timescale: 24000, duration: 0 }
TrackHeaderBox { track_id: 1, disabled: false, duration: 0, width: 0, height: 0 }
MediaHeaderBox { timescale: 24000, duration: 0 }
HandlerBox { handler_type: 1936684398 }
BoxHeader { name: smhd, size: 16, offset: 8 } (skipped)
BoxHeader { name: dinf, size: 36, offset: 8 } (skipped)

2. Add an option in mp4parse_new() to enable log.

3. Print the box name and remain bytes when parsing is aborted. So it prints something like "Dropping 4 bytes in mdhd" in log.

[1]  https://bugzilla.mozilla.org/show_bug.cgi?id=1302027

### DIFF
--- a/mp4parse/src/boxes.rs
+++ b/mp4parse/src/boxes.rs
@@ -33,27 +33,63 @@ macro_rules! box_database {
 
         impl fmt::Debug for BoxType {
             fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-                let box_num: u32 = Into::into(self.clone());
-                write!(f, "{}", convert_to_fourcc(box_num))
+                let fourcc: FourCC = From::from(self.clone());
+                write!(f, "{}", fourcc)
             }
         }
     }
 }
 
-pub fn convert_to_fourcc(number: u32) -> String {
-    let mut box_chars = Vec::new(); 
-    for x in 0..4 {
-        let c = (number >> x * 8 & 0x000000FF) as u8;
-        box_chars.push(c);
+#[derive(Default, PartialEq)]
+pub struct FourCC {
+    pub value: String
+}
+
+impl From<u32> for FourCC {
+    fn from(number: u32) -> FourCC {
+        let mut box_chars = Vec::new();
+        for x in 0..4 {
+            let c = (number >> x * 8 & 0x000000FF) as u8;
+            box_chars.push(c);
+        }
+        box_chars.reverse();
+
+        let box_string = match String::from_utf8(box_chars) {
+            Ok(t) => t,
+            _ => String::from("null"), // error to retrieve fourcc
+        };
+
+        FourCC {
+            value: box_string
+        }
     }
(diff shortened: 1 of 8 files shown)


mp4parse_capi/src/lib.rs
```diff
@@ -248,7 +248,7 @@ impl Read for mp4parse_io {
 
 /// Allocate an `mp4parse_parser*` to read from the supplied `mp4parse_io`.
 #[no_mangle]
-pub unsafe extern fn mp4parse_new(io: *const mp4parse_io) -> *mut mp4parse_parser {
+pub unsafe extern fn mp4parse_new(io: *const mp4parse_io, log: bool) -> *mut mp4parse_parser {
```
rillian: Unless we need logging specifically withing mp4parse_new() I'd rather mirror the rust api and have a separate `mp4parse_log(parser: *mut mp4parse_parser, enable: bool)` call to enable debug output, so we don't have to add the `false` every time we invoke it.

If you want to write a log adaptor which would let us forward trace messages to gecko's logging that would be helpful though. Could be an extra callback in mp4parse_io, or a new method to dump a log buffer.

mp4parse/src/boxes.rs
```diff
@@ -19,7 +20,40 @@ macro_rules! box_database {
...
+        box_chars.push(c);
     }
+    box_chars.reverse();
+
+    let box_string = match String::from_utf8(box_chars) {
+        Ok(t) => t,
+        _ => String::from("error to get 4CC"),
+    };
+
+    box_string
```
rillian: This is fine, but I think it would be better to create a new `Fourcc` type and use that instead of Strings everywhere. You can implement `From<BoxType>` and `From<u32>` to make conversion easy, and we'll have a little more type safety.

We could then write `brands.push(Fourcc::from(be_u32(src)?));` which is a little idiomatic than `convert_to_fourcc()`

mp4parse_capi/src/lib.rs

mp4parse/src/boxes.rs
alfredoyang: Fixed.

mp4parse_capi/src/lib.rs
alfredoyang: Right now I enable rust log when prlog is on for MP4Metadta. I'll think how to forward rust log into gecko's log buffer.

alfredoyang:
Thanks for review. I'll have a follow up PR to use u32 or [u8; 4] instead of String.

PR #55, opened 2017-01-05
parse program_config_element for channel counts

alfredoyang:
This is from [1]. The audio channel count is zero in decoder config descriptor, so we need to calculate the channel count from program_config_element [2].

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1328221
[2] 14496-3  4.4.1 Decoder configuration (GASpecificConfig)

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -1311,45 +1311,26 @@ fn read_ds_descriptor(data: &[u8], esds: &mut ES_Descriptor) -> Result<()> {
         bit_reader.skip(3)?;    // num_assoc_data
         bit_reader.skip(4)?;    // num_valid_cc
 
-        let mono_mixdown_present: u8 = ReadInto::read(bit_reader, 1)?;
-        if mono_mixdown_present > 0 {
+        let mono_mixdown_present: bool = ReadInto::read(bit_reader, 1)?;
+        if mono_mixdown_present {
             bit_reader.skip(4)?;    // mono_mixdown_element_number
         }
 
-        let stereo_mixdown_present: u8 = ReadInto::read(bit_reader, 1)?;
-        if stereo_mixdown_present > 0 {
+        let stereo_mixdown_present: bool = ReadInto::read(bit_reader, 1)?;
+        if stereo_mixdown_present {
             bit_reader.skip(4)?;    // stereo_mixdown_element_number
         }
 
-        let matrix_mixdown_idx_present: u8 = ReadInto::read(bit_reader, 1)?;
-        if matrix_mixdown_idx_present > 0 {
+        let matrix_mixdown_idx_present: bool = ReadInto::read(bit_reader, 1)?;
+        if matrix_mixdown_idx_present {
             bit_reader.skip(2)?;    // matrix_mixdown_idx
             bit_reader.skip(1)?;    // pseudo_surround_enable
         }
 
-        for _ in 0..num_front_channel {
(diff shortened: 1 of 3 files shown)


mp4parse/src/lib.rs
```diff
@@ -1286,7 +1286,71 @@ fn read_ds_descriptor(data: &[u8], esds: &mut ES_Descriptor) -> Result<()> {
...
+        if matrix_mixdown_idx_present > 0 {
+            bit_reader.skip(2)?;    // matrix_mixdown_idx
+            bit_reader.skip(1)?;    // pseudo_surround_enable
+        }
+
+        for _ in 0..num_front_channel {
+            let front_element_is_cpe: u8 = ReadInto::read(bit_reader, 1)?;
+            channel_counts += if front_element_is_cpe > 0 { 2 } else { 1 };
+            bit_reader.skip(4)?;    // front_element_tag_select
+        }
```
rillian: You do this four times. Might be a good candidate for a helper:
```rust
fn read_surround_channel_count(bit_reader: &mut BitReader, channels: u8) -> Result<u16> {
    let mut count = 0;
    for _ in 0..channels {
        let is_cpe: u8 = ReadInto::read(bit_reader, 1)?
        count += if is_cpe > 0 { 2 } else { 1 };
        bit_reader.skip(4)?;
    }
    Ok(count)
}
...
    channel_count += read_surround_channel_count(bit_reader, num_front_channels);
    channel_count += read_surround_channel_count(bit_reader, num_side_channels);
    channel_count += read_surround_channel_count(bit_reader, num_back_channels);
    channel_count += read_surround_channel_count(bit_reader, num_lfe_channels);
```

If you want to be more functional, you can shorten the helper to something like:
```rust
fn read_surround_channel_count(bit_reader: &mut BitReader, channels: u8) -> Result<u16> {
    let count = (0..channels).map(|_| {
        match ReadInto::read(bit_reader, 1)? {
            0 => 1,
            _ => 2,
        }
    }).sum();
    Ok(count)
}
```
I like the `match` better than the `if` but otherwise I don't think either is obviously better.

mp4parse/src/lib.rs
```diff
@@ -1286,7 +1286,71 @@ fn read_ds_descriptor(data: &[u8], esds: &mut ES_Descriptor) -> Result<()> {
...
+        bit_reader.skip(4)?;    // sampling_frequency_index
+        let num_front_channel: u8 = ReadInto::read(bit_reader, 4)?;
+        let num_side_channel: u8 = ReadInto::read(bit_reader, 4)?;
+        let num_back_channel:u8 = ReadInto::read(bit_reader, 4)?;
+        let num_lfe_channel: u8 = ReadInto::read(bit_reader, 2)?;
+        bit_reader.skip(3)?;    // num_assoc_data
+        bit_reader.skip(4)?;    // num_valid_cc
+
+        let mono_mixdown_present: u8 = ReadInto::read(bit_reader, 1)?;
+        if mono_mixdown_present > 0 {
```
rillian: There's an `impl ReadInto for bool` in bitreader 0.2.

alfredoyang: Fixed, thank you for biteader upstream!

mp4parse/src/lib.rs
alfredoyang: Because the bitreader Result is hidden in the functional way, the so I choose the first one. It is unfortunate that Range doesn't support map_err(), otherwise I will rewrite it more functional.

rillian:
Thanks!

PR #59, opened 2017-01-26
create sample table

alfredoyang:
This is the first PR for sample table. There will be following PR to address the TODO items in this patch.

In this PR, it creates a array of mp4parser_indice which mimics the structure in gecko [1]. And pass the array point to gecko. So the memory footprint should be very close to stagefright because stagefright also creates its own Indice nsTArray.

There are 2 TODO items, 'elst' and 'ctts', I believe parsing 'elst' is more priority because it has the precise sample composition time.

I also try to make this patch more functional way for a while. But it looks like I probably don't get the point of functional program yet.

[1] http://searchfox.org/mozilla-central/rev/bf98cd4315b5efa1b28831001ad27d54df7bbb68/media/libstagefright/binding/include/mp4_demuxer/Index.h#46

### DIFF
--- a/mp4parse_capi/src/lib.rs
+++ b/mp4parse_capi/src/lib.rs
@@ -116,7 +116,7 @@ pub struct mp4parse_track_info {
 
 #[repr(C)]
 #[derive(Default, Debug)]
-pub struct mp4parser_indice {
+pub struct mp4parse_indice {
     pub start_offset: u64,
     pub end_offset: u64,
     pub start_composition: u64,
@@ -128,17 +128,17 @@ pub struct mp4parser_indice {
 #[repr(C)]
 pub struct mp4parse_byte_data {
     pub length: u32,
-    // cheddar can't handle gereric type, so it needs to be multiple data types here.
+    // cheddar can't handle generic type, so it needs to be multiple data types here.
     pub data: *const u8,
-    pub indices: *const mp4parser_indice,
+    pub indices: *const mp4parse_indice,
 }
 
 impl Default for mp4parse_byte_data {
     fn default() -> Self {
         mp4parse_byte_data {
             length: 0,
-            data: std::ptr::null_mut(),
-            indices: std::ptr::null_mut(),
+            data: std::ptr::null(),
+            indices: std::ptr::null(),
         }
     }
 }
@@ -148,7 +148,7 @@ impl mp4parse_byte_data {
         self.length = data.len() as u32;
         self.data = data.as_ptr();
     }
-    fn set_indices(&mut self, data: &Vec<mp4parser_indice>) {
+    fn set_indices(&mut self, data: &Vec<mp4parse_indice>) {
         self.length = data.len() as u32;
         self.indices = data.as_ptr();
     }
@@ -162,7 +162,7 @@ pub struct mp4parse_pssh_info {
 
 #[repr(C)]
 #[derive(Default)]
-pub struct mp4parser_sinf_info {
+pub struct mp4parse_sinf_info {
     pub is_encrypted: u32,
     pub iv_size: u8,
     pub kid: mp4parse_byte_data,
@@ -176,7 +176,7 @@ pub struct mp4parse_track_audio_info {
     pub sample_rate: u32,
     pub profile: u16,
     pub codec_specific_config: mp4parse_byte_data,
-    pub protected_data: mp4parser_sinf_info,
+    pub protected_data: mp4parse_sinf_info,
 }
 
 #[repr(C)]
@@ -187,7 +187,7 @@ pub struct mp4parse_track_video_info {
     pub image_width: u16,
     pub image_height: u16,
     pub extra_data: mp4parse_byte_data,
-    pub protected_data: mp4parser_sinf_info,
+    pub protected_data: mp4parse_sinf_info,
 }
 
 #[repr(C)]
@@ -206,7 +206,7 @@ struct Wrap {
     poisoned: bool,
     opus_header: HashMap<u32, Vec<u8>>,
     pssh_data: Vec<u8>,
-    sample_table: HashMap<u32, Vec<mp4parser_indice>>,
+    sample_table: HashMap<u32, Vec<mp4parse_indice>>,
 }
 
 #[repr(C)]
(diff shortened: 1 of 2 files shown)


mp4parse_capi/src/lib.rs
```diff
@@ -615,6 +640,149 @@ pub unsafe extern fn mp4parse_get_track_video_info(parser: *mut mp4parse_parser,
...
+                                start_decode: 0,
+                                sync: !has_sync_table,
+                            }
+                        );
+                    }
+                }
+            true
+        });
+    if !ret {
+        return None;
```
rillian: Since this check is so far from the declaration, please use a more descriptive variable name than `ret`. Maybe `have_table_entries` or `table_valid`, something like that, to be clear why you can't just check `sample_table.is_empty()`.

mp4parse_capi/src/lib.rs
```diff
@@ -114,16 +115,30 @@ pub struct mp4parse_track_info {
...
+    pub start_composition: u64,
+    pub end_composition: u64,
+    pub start_decode: u64,
+    pub sync: bool,
+}
+
+#[repr(C)]
 pub struct mp4parse_byte_data {
     pub length: u32,
+    // cheddar can't handle gereric type, so it needs to be multiple data types here.
```
rillian: s/gereric/generic/

mp4parse_capi/src/lib.rs
```diff
@@ -114,16 +115,30 @@ pub struct mp4parse_track_info {
 }
 
 #[repr(C)]
+#[derive(Default, Debug)]
+pub struct mp4parser_indice {
```
rillian: Please use `mp4parse_indice` for consistency with other structs. "mp4parser_noun" is more grammatical, but we can ignore that treating "mp4parse" as a prefix instead of an adjective. :)

mp4parse_capi/src/lib.rs
```diff
@@ -114,16 +115,30 @@ pub struct mp4parse_track_info {
...
     pub data: *const u8,
+    pub indices: *const mp4parser_indice,
 }
 
 impl Default for mp4parse_byte_data {
     fn default() -> Self {
         mp4parse_byte_data {
             length: 0,
             data: std::ptr::null_mut(),
+            indices: std::ptr::null_mut(),
```
rillian: Hmm. I think these should both be `std::ptr::null()`. This works because `*mut` coerces to `*const` which the struct has in both cases.

My mistake. Should probably change them both here now that we know.

mp4parse_capi/src/lib.rs
```diff
@@ -615,6 +640,149 @@ pub unsafe extern fn mp4parse_get_track_video_info(parser: *mut mp4parse_parser,
...
+                }
+            true
+        });
+    if !ret {
+        return None;
+    }
+
+    // Mark the sync sample in sample_table according to 'stss'.
+    let stss_iter = stss.samples.iter();
+    for iter in stss_iter {
```
rillian: `for iter in &stss.samples { ... }` ?

alfredoyang:
Thanks for review. I'll add testcase later.

mp4parse_capi/src/lib.rs
```diff
@@ -741,13 +741,12 @@ fn create_sample_table(track: &Track) -> Option<Vec<mp4parser_indice>> {
                 }
             true
         });
-    if !ret {
+    if sample_table.is_empty() {
```
rillian: I just wanted to confirm if this this what you want? `ret` carried the return value of `all()` so it was false if any `end_offset` was zero, with `is_empty()` is false if any `end_offset` *wasn't* zero. Effectively converting the `.all()` into an `.any()`.

If this is want you want, I suggest using a second `.map()` and moving the `cur_position` update and `sample_table.push()` into the conditional so the logic is simpler.

alfredoyang: It's my bad without explaining it early. My original idea is returning false if any of ```end_offset``` was zero. So when anything wrong in ```stsz.sample_sizes``` or ```stsc.samples```, we can identify it quickly.

I'm not quite sure your second ```.map()```, do you mean codes like this?
```
    stsc.samples.iter()
        .map(|sample_chunk| {
                // Convert the compact table to chunk range and samples per trunk.
                start_chunk = end_chunk;
                end_chunk += sample_chunk.first_chunk;
                ((start_chunk .. end_chunk), sample_chunk.samples_per_chunk)
            })
        .map(|chunks| {
                // Chunks in the chunk range have the same sample number.
                let chunks_range = chunks.0;
                let samples_per_chunk = chunks.1;
                for chunk_id in chunks_range {
                    let mut cur_position: u64 = stco.offsets[chunk_id as usize];
                    // Create sample according to sample number of each trunk and calculate
                    // sample offset.
                    for _ in 0 .. samples_per_chunk {
                        let start_offset = cur_position;
                        let end_offset = match (stsz.sample_size, sample_size_iter.next()) {
                            (_, Some(t)) => start_offset + *t as u64,
                            (t, _) if t > 0 => start_offset + t as u64,
                            _ => 0,
                        };
                        if end_offset > 0 {
                            cur_position = end_offset;

                            sample_table.push(
                                mp4parse_indice {
                                    start_offset: start_offset,
                                    end_offset: end_offset,
                                    start_composition: 0,
                                    end_composition: 0,
                                    start_decode: 0,
                                    sync: !has_sync_table,
                                }
                            );
                        }
                    }
                }
        });

    if sample_table.is_empty() {
        return None;
    }
```

rillian: Yes, that's what I was thinking.

alfredoyang: hmm...
It seems ```.map()``` is lazy, I guess we need to add something like ```.colletc()``` or ```.sum()``` after ```.map()```?

```
warning: unused result which must be used: iterator adaptors are lazy and do nothing unless consumed, #[warn(unused_must_use)] on by default
   --> src/lib.rs:715:5
    |
715 |     stsc.samples.iter()
    |     ^
```

rillian: Oops, you're right! I worry about `.collect()` being inefficient when we don't need the resulting vector. `.count()` might be more semantically-neutral than `.all()` or `.any()`.

I asked on irc; #rust-beginners recommends using a for loop instead of map().something() when the primary result is a side-effect.

alfredoyang: OK, I change it to a for loop. I think it's probably better to use ```.map()``` to where nothing could be wrong.

PR #61, opened 2017-02-08
fix sample table and add testcase

alfredoyang:
1. Add testcase.
2. Fix some issues in previous PR including: 'stss' should be optional table; the chunk_table is wrong; track sample time is wrong.


mp4parse_capi/src/lib.rs
```diff
@@ -683,9 +683,9 @@ fn create_sample_table(track: &Track) -> Option<Vec<mp4parse_indice>> {
         _ => return None,
     };
 
-    let (stsc, stco, stsz, stss, stts) =
```
alfredoyang: 'stss' is optional box according to spec.

mp4parse_capi/src/lib.rs
```diff
@@ -698,17 +698,22 @@ fn create_sample_table(track: &Track) -> Option<Vec<mp4parse_indice>> {
...
     let mut sample_size_iter = stsz.sample_sizes.iter();
-    let mut start_chunk = 0;
-    let mut end_chunk = 0;
-    let chunk_iter = stsc.samples.iter()
-        .map(|sample_chunk| {
-                // Convert the compact table to chunk range and samples per trunk.
-                start_chunk = end_chunk;
-                end_chunk += sample_chunk.first_chunk;
-                ((start_chunk .. end_chunk), sample_chunk.samples_per_chunk)
-            });
```
alfredoyang: We should calculate in reverse order so we can convert a the compact table to correct chunk range.

mp4parse_capi/src/lib.rs
```diff
@@ -743,8 +748,13 @@ fn create_sample_table(track: &Track) -> Option<Vec<mp4parse_indice>> {
     }
 
     // Mark the sync sample in sample_table according to 'stss'.
-    for iter in &stss.samples {
-        sample_table[(iter - 1) as usize].sync = true;
```
alfredoyang: Obviously, my previous test stream is not good enough, so these problems doesn't be found.

alfredoyang:
Hmm... I'm re-factoring the whole function so the patch could be more readable. It's probably better to close this for now.

PR #64, opened 2017-02-15
refactor sample table and support elst/ctts tables

alfredoyang:
1. Refactor deep for-loop by iterator, I hope it makes this patch more readable.
2. Support 'ctts' box in parser.
3. Support 'elst' box in capi.
4. Add testcase.


mp4parse_capi/src/lib.rs
```diff
@@ -677,15 +695,172 @@ pub unsafe extern fn mp4parse_get_indice_table(parser: *mut mp4parse_parser, tra
...
+    fn next_offset_time(&mut self) -> i64 {
+        match self.next() {
+            Some(v) => v as i64,
+            _ => 0,
+        }
+    }
+}
+
+// Convert 'stts' compact table to full table by iterator,
+// (smaple_count_with_the_same_time, time) => (time, time, time) ... repeats
```
rillian: s/smaple/sample/

mp4parse_capi/src/lib.rs
```diff
@@ -677,15 +695,172 @@ pub unsafe extern fn mp4parse_get_indice_table(parser: *mut mp4parse_parser, tra
...
+    }
+}
+
+// Convert 'stco' compact table to full table by iterator.
+// (start_chunk_num, sample_number) => (sample_number), (sample_number) .. repeat
+// (next start_chunk_num - current start_chunk_num).
+//
+// For example:
+// (1, 5), (5, 10), (9, 2) => (5, 5, 5, 5, 10, 10, 10, 10, 2)
+struct SampleToChunkIterator<'a> {
```
rillian: The example here isn't correct. The Iterator returns (id, count) tuples.

alfredoyang:
Thanks foe comments. I'll add ````cslg```` as TODO in following PR.

mp4parse/src/lib.rs
```diff
@@ -194,6 +194,17 @@ pub struct Sample {
     pub sample_delta: u32,
 }
 
+#[derive(Debug, Clone)]
+pub struct TimeOffset {
+    pub sample_count: u32,
+    pub time_offset: i32,
```

mp4parse/src/lib.rs
```diff
@@ -1128,6 +1145,37 @@ fn read_stsc<T: Read>(src: &mut BMFFBox<T>) -> Result<SampleToChunkBox> {
     })
 }
 
+fn read_ctts<T: Read>(src: &mut BMFFBox<T>) -> Result<CompositionOffsetBox> {
+    let (version, _) = read_fullbox_extra(src)?;
+
+    let counts = be_u32(src)?;
+    let mut offsets = Vec::new();
+    for _ in 0..counts {
```

mp4parse/src/lib.rs
```diff
@@ -1128,6 +1145,37 @@ fn read_stsc<T: Read>(src: &mut BMFFBox<T>) -> Result<SampleToChunkBox> {
...
+        let (sample_count, time_offset) = match version {
+            0 => {
+                let count = be_u32(src)?;
+                let offset = be_u32(src)? as i32;
+                (count, offset)
+            },
+            _ => {
```

mp4parse_capi/src/lib.rs
```diff
@@ -665,7 +665,25 @@ pub unsafe extern fn mp4parse_get_indice_table(parser: *mut mp4parse_parser, tra
...
+            },
+            (&Some(empty_duration), _, &Some(scale)) => {
+                (empty_duration.0 as i64, scale.0 as i64)
+            },
+            (_, &Some(media_time), &Some(scale)) => {
+                ((0 - media_time.0) as i64, scale.0 as i64)
+            },
+            _ => (0, 0)
+        };
+    offset_time *= scale;
```

mp4parse_capi/src/lib.rs
```diff
@@ -677,15 +695,172 @@ pub unsafe extern fn mp4parse_get_indice_table(parser: *mut mp4parse_parser, tra
     MP4PARSE_ERROR_INVALID
 }
 
-fn create_sample_table(track: &Track) -> Option<Vec<mp4parse_indice>> {
+// Convert a 'ctt' compact table to full table by iterator,
```

mp4parse_capi/src/lib.rs
```diff
@@ -677,15 +695,172 @@ pub unsafe extern fn mp4parse_get_indice_table(parser: *mut mp4parse_parser, tra
...
+// (sample_with_the_same_offset_count, offset) => (offset), (offset), (offset) ...
+//
+// For example:
+// (2, 10), (4, 9) into (10, 10, 9, 9, 9, 9) by calling next_offset_time().
+struct TimeOffsetIteraor<'a> {
```

mp4parse_capi/src/lib.rs

mp4parse_capi/src/lib.rs

mp4parse/src/lib.rs
```diff
@@ -1128,6 +1151,43 @@ fn read_stsc<T: Read>(src: &mut BMFFBox<T>) -> Result<SampleToChunkBox> {
...
 
+fn read_ctts<T: Read>(src: &mut BMFFBox<T>) -> Result<CompositionOffsetBox> {
+    let (version, _) = read_fullbox_extra(src)?;
+
+    let counts = be_u32(src)?;
+    let mut offsets = Vec::new();
+    for _ in 0..counts {
+        if src.bytes_left() == 0 {
+            return Err(Error::InvalidData("invalid data in 'ctts' box"));
+        }
```
rillian: I meant that you know the size of the TimeOffset table entries, so you can calculate how many bytes will be read before entering the loop. That way you can return early.

mp4parse_capi/src/lib.rs
```diff
@@ -677,15 +694,186 @@ pub unsafe extern fn mp4parse_get_indice_table(parser: *mut mp4parse_parser, tra
...
+                    _ => (0 .. 0),
+                };
+
+                self.cur_sample_count.next()
+            });
+
+        match has_sample {
+            Some(_) => Some(self.cur_sample_delta),
+            _ => None,
+        }
```
rillian: FWIW, you can do this in one line with `has_sample.and(Some(self.cur_sample_delta))` but the `match` may be more clear. Up to you.

mp4parse_capi/src/lib.rs
```diff
@@ -695,84 +883,123 @@ fn create_sample_table(track: &Track) -> Option<Vec<mp4parse_indice>> {
...
+        sort_table.sort_by_key(|i| {
+            match sample_table.get(*i) {
+                Some(v) => {
+                    v.start_composition
+                },
+                _ => 0,
             }
+        });
+
+        let iter = sort_table.iter().peekable();
```
rillian: Do you need to call `.peekable()` here? You never call `iter.peek()` and the length should be the same for both Iterators.

mp4parse/src/lib.rs

mp4parse_capi/src/lib.rs
alfredoyang: Thanks!
I like to reduce the lines. :-)

mp4parse_capi/src/lib.rs
alfredoyang: Urghh... I forgot to remove it during the re-arch.

mp4parse/src/lib.rs
```diff
@@ -1155,11 +1155,13 @@ fn read_ctts<T: Read>(src: &mut BMFFBox<T>) -> Result<CompositionOffsetBox> {
     let (version, _) = read_fullbox_extra(src)?;
 
     let counts = be_u32(src)?;
+
+    if src.bytes_left() < (counts as usize * std::mem::size_of::<TimeOffsetVersion>()) {
```
rillian: Thanks for hoisting this. I think `(counts as usize * 8)` would be better here. One could add new fields to `TimeOffsetVersion` and change this calculation without changing the number of bytes we're reading.

alfredoyang: Sure, I'll update it.

alfredoyang:
Oh, I included an another fix by accident. So there are some removed codes here.

PR #70, opened 2017-02-28
Verify we're compiled with runtime overflow checking.

rillian:
The library assumes runtime checking of integer arithmetic
overflows, which was much nicer than putting checked_foo()
methods everywhere. We set debug-assertions = true in
Cargo.toml to enable this even for release builds, where
the compiler normally turns it off for performance reasons.

However, it's not clear this is working with rust 1.15.1.
Add a test which checks for the expected panic.

### DIFF
--- a/mp4parse/tests/overflow.rs
+++ b/mp4parse/tests/overflow.rs
@@ -1,0 +1,15 @@
+/// Verify we're built with run-time integer overflow detection.
+
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#[test]
+#[should_panic(expected = "attempt to add with overflow")]
+fn overflow_protection() {
+    let edge = u32::max_value();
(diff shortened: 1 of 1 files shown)


kinetiknz:
Related: https://bugzilla.mozilla.org/show_bug.cgi?id=1342848

nox:
If I understand correctly, this means `mp4parse` can panic on illformed data, right? Shouldn't it instead use checked arithmetic explicitly and return errors?

kinetiknz:
Yes, the ideal is to use checked arithmetic and return errors rather than panic on malformed data.

Unfortunately, it's possible to forget to do that, and rather than leave the burden entirely on the writer/reviewer, we'd prefer to turn such mistakes into fatal errors rather than silently computing invalid values (in release builds).  It might not matter as much in pure Rust code, but since our primary use case is exporting this data to Gecko via a C FFI we'd rather not allow the elevation of overflow errors into potential security bugs on the C side.

nox:
There is nothing wrong about strengthening `mp4parse_capi` with regard to panics, but `mp4parse` shouldn't rely on panics to signal errors.

PR #74, opened 2017-03-03
Fix negative composition time

alfredoyang:
In [1], the composition time could be negative due to edit list in ```elst```. So it changes the u64 to i64 to handle this case.
We also fixed the wrong media_time unit, it should be track based unit, not the scale in ```mdhd```.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1343793


mp4parse_capi/tests/test_sample_table.rs
```diff
@@ -95,3 +95,59 @@ fn parse_sample_table() {
...
+            track_id: 0,
+            duration: 0,
+            media_time: 0,
+        };
+        rv = mp4parse_get_track_info(parser, 1, &mut track_info);
+        assert_eq!(rv, mp4parse_error::MP4PARSE_OK);
+        assert_eq!(track_info.track_type, mp4parse_track_type::MP4PARSE_TRACK_TYPE_AUDIO);
+        assert_eq!(track_info.codec, mp4parse_codec::MP4PARSE_CODEC_AAC);
+
+        // Check audio smaple table
```
rillian: s/smaple/sample/ and period at the end, please.

mp4parse_capi/tests/test_sample_table.rs
```diff
@@ -95,3 +95,59 @@ fn parse_sample_table() {
...
+        let mut is_fragmented_file: u8 = 0;
```
rillian: The test would be more robust if you initialize to an impossible value, like `2u8` to verify the method is setting the expected value, which is also zero in this case.

mp4parse_capi/src/lib.rs
```diff
@@ -668,19 +668,28 @@ pub unsafe extern fn mp4parse_get_indice_table(parser: *mut mp4parse_parser, tra
         _ => {},
     }
 
+    // track.media_time is track based time unit.
+    let media_time = match (&track.media_time, &track.timescale) {
+        (&Some(t), &Some(s)) => {
+            let time_us = PresentationTime::new(t.0 as i64, s);
```
rillian: Better to call this just `time`. It hasn't been converted to µs yet.

mp4parse_capi/src/lib.rs
```diff
@@ -668,19 +668,28 @@ pub unsafe extern fn mp4parse_get_indice_table(parser: *mut mp4parse_parser, tra
...
             },
-            (&Some(empty_duration), _, &Some(scale)) => {
-                empty_duration.0 as i64 * scale.0 as i64
+            (&Some(empty_duration), &Some(scale), _) => {
+                (empty_duration.0 * scale.0) as i64
             },
-            (_, &Some(media_time), &Some(scale)) => {
-                (0 - media_time.0 as i64) * scale.0 as i64
+            (_, _, Some(media_time)) => {
+                media_time
```
rillian: It looks like this is changing the sign relative to the previous patch. Just wanted to make sure that's what you intend.

mp4parse_capi/tests/test_sample_table.rs
alfredoyang: fixed.

mp4parse_capi/tests/test_sample_table.rs
alfredoyang: Fixed.

mp4parse_capi/src/lib.rs

mp4parse_capi/src/lib.rs
alfredoyang: It is not relative to previous change. media_time declares as i64, so we don't need to cast it to i64 again.

The media_time should be scaled by ```mvhd.timescale```, not ```mdhd.timescale```. My original implementation is wrong.

kinetiknz: Timestamps should be stored in MediaScaledTime or TrackScaledTime (as appropriate) to avoid this kind of bug, we should never as raw {i,u}64.

alfredoyang: Yes...
Is that better to implement a convert function in TrackScaledTime/MediaScaledTime? So we can avoid this kind of error again?

kinetiknz: There's media_time_to_us and track_time_to_us for that, unless you mean a different conversion?

alfredoyang: See. I'll update it in this PR.
Thank you!

PR #77, opened 2017-03-09
handle ctts negative value

alfredoyang:
1. Always use Version1 when parsing ```ctts``` due to content problem.
2. Handle ```ctts``` negative value when calculating composition time.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1345382


mp4parse_capi/src/lib.rs
```diff
@@ -964,18 +950,22 @@ fn create_sample_table(track: &Track, track_offset_time: i64) -> Option<Vec<mp4p
...
+        sum_delta += stts_iter.next_delta() as i64;
 
         // ctts_offset is the current sample offset time.
-        let ctts_offset = PresentationTime::new(ctts_offset_iter.next_offset_time(), timescale);
+        let ctts_offset = ctts_offset_iter.next_offset_time();
 
-        let start_composition = decode_time + ctts_offset.to_us() + track_offset_time;
-        let end_composition = sum_delta.to_us() + ctts_offset.to_us() + track_offset_time;
+        // ctts_offset could be negative but (decode_time + ctts_offset) should always be positive
+        // value.
```
rillian: Better to verify this condition here where you can return a None, I think?

mp4parse_capi/src/lib.rs
```diff
@@ -393,6 +393,15 @@ fn track_time_to_us(time: TrackScaledTime, scale: TrackTimeScale) -> Option<u64>
     rational_scale(time.0, scale.0, microseconds_per_second)
 }
 
+fn get_track_time_to_us(time: i64, scale: TrackTimeScale) -> i64 {
+    assert!(time >= 0);
+    let track_time = TrackScaledTime(time as u64, scale.1);
```
rillian: This seems messy. I guess taking a specific type here preserves some type checking of the scale value, but it would be neater to keep a specific type for the track time itself. Can you make encapsulate the signed scaling in another type, perhaps using a `rational_scale_signed()` or making `rational_scale` generic? I'm not sure if there's a simple trait bound for integer types.

alfredoyang: I plan to use something like [1]. We may also need to make TrackScaledTime generic for both i64/u64, I think.

[1] http://stackoverflow.com/questions/36013519/how-can-i-make-a-rust-function-accept-any-floating-type-as-an-argument

mp4parse_capi/src/lib.rs
alfredoyang: Yes, sure.

mp4parse_capi/src/lib.rs
```diff
@@ -779,6 +787,7 @@ struct TimeToSampleIteraor<'a> {
     cur_sample_count: std::ops::Range<u32>,
     cur_sample_delta: u32,
     stts_iter: std::slice::Iter<'a, mp4parse::Sample>,
+    track_id: usize,
 }
 
 impl<'a> Iterator for TimeToSampleIteraor<'a> {
```
kinetiknz: Can you please fix the spelling of Iterator since you're touching this area?

mp4parse_capi/src/lib.rs
```diff
@@ -1408,7 +1414,7 @@ fn media_time_overflow() {
 
 #[test]
 fn track_time_overflow() {
-  let scale = TrackTimeScale(44100, 0);
-  let duration = TrackScaledTime(4413527634807900, 0);
+  let scale = TrackTimeScale(44100 as u64, 0);
```
kinetiknz: 44100u64 is slightly better maybe

mp4parse_capi/src/lib.rs

mp4parse_capi/src/lib.rs

alfredoyang:
Thanks for review!

PR #79, opened 2017-03-16
get audio codec specific data

alfredoyang:
In Gecko, audio codec specific data (descriptor tag 0x05 in ```esds```) is required. PlaformDecoderModule like ffmpeg needs this field to initialize decoder.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1347834

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -1413,6 +1413,7 @@ fn read_ds_descriptor(data: &[u8], esds: &mut ES_Descriptor) -> Result<()> {
     esds.audio_object_type = Some(audio_object_type);
     esds.audio_sample_rate = sample_frequency;
     esds.audio_channel_count = Some(channel_counts);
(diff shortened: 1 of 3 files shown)


mp4parse/src/lib.rs
```diff
@@ -1412,6 +1413,7 @@ fn read_ds_descriptor(data: &[u8], esds: &mut ES_Descriptor) -> Result<()> {
     esds.audio_object_type = Some(audio_object_type);
     esds.audio_sample_rate = sample_frequency;
     esds.audio_channel_count = Some(channel_counts);
+    esds.decoder_specific_data.extend(data.iter());
```
kinetiknz: You're setting the entire contents of data here, not adding to it, right? Is it worth asserting that decoder_specific_data is empty before calling extend?

alfredoyang: yeah, I'll add it.

PR #81, opened 2017-03-30
Fix UB caused by casting a Rust fn ptr to a *c_void (BMO bug #1351497)

kinetiknz:
Use Option<extern fn()> to represent potentially-null function pointer.
    
This addresses https://bugzilla.mozilla.org/show_bug.cgi?id=1351497, which was filed as a rustc bug in https://github.com/rust-lang/rust/issues/40913, but revealed to be undefined behaviour that happened to work with earlier compiler versions.
    
As the comment removed in this patch describes, the only reason this code existed was to work around a limitation of rusty-cheddar.

Not sure what to do about the rusty-cheddar issue.  We can patch the generated header manually (by replacing the bogus mp4parse_io member `Option read` with `intptr_t (*read)(uint8_t* buffer, uintptr_t size, void* userdata)` for now, but that's not an ideal solution.

### DIFF
--- a/mp4parse_capi/Cargo.toml
+++ b/mp4parse_capi/Cargo.toml
@@ -29,7 +29,7 @@ mp4parse = {version = "0.7.1", path = "../mp4parse"}
 num-traits = "0.1.37"
 
 [build-dependencies]
-rusty-cheddar = "0.3.2"
+rusty-cheddar = { git = "https://github.com/kinetiknz/rusty-cheddar" }
 
 [features]
 fuzz = ["mp4parse/fuzz"]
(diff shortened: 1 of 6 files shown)


rillian:
I don't want to go back to maintaining `mp4parse.h` by hand, so I think we need to fix cheddar if we're going to use `Option<extern fn ...>`. Or move from cheddar to bindgen. :/

kinetiknz:
Okay, so I've forked rusty-cheddar and fixed it: https://github.com/kinetiknz/rusty-cheddar/commit/bb819e501c9031ce62e409029c4ea5b2c736b633

And repushed this fix with the rusty-cheddar dep switched to my repo.

rillian:
Most awesome, thanks!

PR #84, opened 2017-04-04
Migrate to moz-cheddar 0.4.0.

rillian:
Move to our published fork of rusty-cheddar instead of
building against a git repo.

### DIFF
--- a/mp4parse_capi/Cargo.toml
+++ b/mp4parse_capi/Cargo.toml
@@ -29,7 +29,7 @@ mp4parse = {version = "0.7.1", path = "../mp4parse"}
 num-traits = "0.1.37"
 
 [build-dependencies]
-rusty-cheddar = { git = "https://github.com/kinetiknz/rusty-cheddar" }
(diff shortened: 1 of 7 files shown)


rillian:
This needs updating for the enum change.

rillian:
I'm not sure about `MP4PARSE::OK` vs `mp4parse_error::OK`. Lowercase typename is more clear.

kinetiknz:
> I'm not sure about MP4PARSE::OK vs mp4parse_error::OK. Lowercase typename is more clear.

Yeah, it's a bit of a pain because you want (in this case, anyway) a lowercase enum type name but uppercase enum member names.  We could change moz-cheddar to give us what we want -- it might be a bit specialized, but it's probably not worth worrying about trying to be too general... ultimately it's missing a bunch of control over the naming of generated stuff.

MP4PARSE as an error enum typename seems a bit awkward, but I guess mp4parse_error::OK is too.  I slightly lean towards the latter, but it's not a big deal.

PR #89, opened 2017-04-13
Remove empty_sample_boxes.

alfredoyang:
Fragmented file can be recognized from stsc, stco, and stts. There is no requirement to keep empty_sample_boxes anymore.

### DIFF
--- a/mp4parse_capi/src/lib.rs
+++ b/mp4parse_capi/src/lib.rs
@@ -1051,9 +1051,8 @@ pub unsafe extern fn mp4parse_is_fragmented(parser: *mut mp4parse_parser, track_
     let mut iter = tracks.iter();
     iter.find(|track| track.track_id == Some(track_id)).map_or(mp4parse_status::BAD_ARG, |track| {
         match (&track.stsc, &track.stco, &track.stts) {
-            (&Some(ref stsc), &Some(ref stco), &Some(ref stts)) if stsc.samples.is_empty() &&
-                                                                   stco.offsets.is_empty() &&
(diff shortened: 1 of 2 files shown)


alfredoyang:
hmm... Travis CI fails at rust nightly build?

rillian:
The nightly failure is an issue with rustup and the current nightly packaging. See https://github.com/rust-lang-nursery/rustup.rs/issues/1062

mp4parse_capi/src/lib.rs
```diff
@@ -1049,13 +1049,15 @@ pub unsafe extern fn mp4parse_is_fragmented(parser: *mut mp4parse_parser, track_
...
-    }
-
-    mp4parse_status::OK
+    iter.find(|track| track.track_id == Some(track_id)).map_or(mp4parse_status::BAD_ARG, |track| {
+        match (&track.stsc, &track.stco, &track.stts) {
+            (&Some(ref stsc), &Some(ref stco), &Some(ref stts)) if stsc.samples.is_empty() &&
+                                                                   stco.offsets.is_empty() &&
+                                                                   stts.samples.is_empty() => (*fragmented) = true as u8,
+            _ => {},
+        };
```
rillian: I think moving all this code to a single location is a good idea. And your implementation here is marvelously functional, but the lines are also very long. I tried several different forms, but couldn't find a better way to handle the nested Options. Maybe just break the line before the `if` guard in the match?

rillian: yes, that's better I think.

alfredoyang: Thanks, PR updated.

Issue #90, opened 2017-04-13
fail to parse invalid init segment


rillian:
From this [gecko bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1356291) we fail to parse the init segment from http://dash.edgesuite.net/envivio/Envivio-dash2/v4_258-Header.m4s with "unexpected EOF".

This is invalid, but stagefright appears to handle it.

alfredoyang:
Something happens in ```read_es_descriptor()```, checking now.

alfredoyang:
Ok, I think I found the problem, followings are its esds data:

```
000001d0: 00bb 8000 0000 0000 2a65 7364 7300 0000 ........*esds...
000001e0: 0003 801b 0000 0004 8012 4015 0006 0000 ..........@.....
000001f0: 01fe f000 0000 0005 8002 1190 0601 0200 ................
00000200: 0000 1073 7474 7300 0000 0000 0000 0000 ...stts.........
```

The es_descriptor data is ```0003 801b 0000 00```, ```80``` is the **optional extended descriptor type tag**.

The optional extended descriptor type should be three bytes ```80 80 80```; however, there is only one byte in this sample. So it causes a problem in https://github.com/mozilla/mp4parse-rust/blob/master/mp4parse/src/lib.rs#L1331.

I can't find spec mention about this [1] but stagefright just ignores the ```80``` regardless how many bytes it has. So here I'll file a PR which just ignores the ```80``` up to three bytes.

[1] 14496-1 7.3.3.2

jyavenard:
if we make no use of that box content, why not skip it entirely?

alfredoyang:
That's what we do, but we need to calculate the length of es_descriptor to skip it. In this case, the one byte ```80``` optional extended descriptor causes wrong length calculation.

alfredoyang:
https://github.com/mozilla/mp4parse-rust/pull/91

jyavenard:
I don't understand why you would need to read the content to calculate it size. The size that followed the box type is all that needed no?

Unless the box size is wrong, but if that was the case stage fright wouldn't read this file either..

alfredoyang:
Because dc_descriptor is inside es_descriptor, you can't know dc_descriptor location without parsing es_descriptor.

jyavenard:
Sure. But do we care about dc_descriptor?

alfredoyang:
[...]

PR #91, opened 2017-04-15
fix for one byte extend descriptor tag

alfredoyang:
This is for https://github.com/mozilla/mp4parse-rust/issues/90.

Optional extended descriptor type tag size is one byte in this case, not three. Instead of always checking three bytes, this PR checks the extension descriptor tag from one up to three bytes to handle this case.


jyavenard:
I think this is the wrong approach, we shouldn't read the content of this box at all, just skip it.

alfredoyang:
Hi jya, please check my reply in https://github.com/mozilla/mp4parse-rust/issues/90.

Issue #93, opened 2017-05-17
Support rev 1.0 vpcc box


rillian:
Since FFmpeg's commit 64ad44a381 (https://github.com/FFmpeg/FFmpeg/commit/64ad44a381), Firefox no longer plays vp9 in mp4 files.

This commit updates the vpcc box to be per 1.0 spec:

https://github.com/webmproject/vp9-dash/blob/master/VPCodecISOMediaFileFormatBinding.md

(from https://bugzilla.mozilla.org/show_bug.cgi?id=1365787)

est31:
Can this be closed now?

kinetiknz:
Yeah, thanks for pointing it out.  Done in #96.

PR #96, opened 2017-05-23
Update vpcc parsing to 'VP Codec ISO Media File Format' version 1

alfredoyang:
#93

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -301,6 +301,7 @@ pub struct VPxConfigBox {
     pub color_space: u8, // Really an enum
     pub chroma_subsampling: u8,
     transfer_function: u8,
+    matrix: Option<u8>, // Available in 'VP Codec ISO Media File Format' version 1 only.
     video_full_range: bool,
     pub codec_init: Vec<u8>, // Empty for vp8/vp9.
 }
@@ -403,6 +404,7 @@ pub enum CodecType {
     Opus,
     H264,   // 14496-10
     MP4V,   // 14496-2
+    VP10,
     VP9,
     VP8,
     EncryptedVideo,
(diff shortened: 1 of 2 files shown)


mp4parse/src/tests.rs
```diff
@@ -388,6 +388,36 @@ fn read_vpcc() {
...
+#[test]
+fn read_vpcc_version_1() {
+    let data_length = 12u16;
+    let mut stream = make_fullbox(BoxSize::Auto, b"vpcC", 1, |s| {
+        s.B8(2)     // profile
+         .B8(0)     // level
+         .B8(0x82)  // bitdepth (4 bits), chroma (3 bits), video full range (1 bit)
+         .B8(0)     // color primaries
+         .B8(0)     // transfer characteristics
+         .B8(0)     // matrix
```
rillian: The draft spec says, "If `matrixCoefficients` is 0 (RGB), the chroma sampling MUST be 3 (4:4:4)." So I think you need to set whatever the YCbCr value is for matrix, or set 0b1000_011_0 for the bitdepth/chroma/range field.

mp4parse/src/boxes.rs
```diff
@@ -120,6 +120,7 @@ box_database!(
     ESDBox                            0x65736473, // "esds"
     VP8SampleEntry                    0x76703038, // "vp08"
     VP9SampleEntry                    0x76703039, // "vp09"
+    VP10SampleEntry                   0x76703130, // "vp10"
```
rillian: It's tempting to ignore the VP10 variant, since the spec doesn't define anything about it beyond the name, and we're not planning to support it in Firefox. I suppose it does no harm though.

mp4parse/src/tests.rs
rillian: ISO 23001-8 is [available](http://standards.iso.org/ittf/PubliclyAvailableStandards/index.html) without charge. It looks like `1` (BT.709) would work fine for the matrix field.

rillian: Color primaries should be 1 as well.

mp4parse/src/lib.rs
```diff
@@ -1250,20 +1252,34 @@ fn read_stts<T: Read>(src: &mut BMFFBox<T>) -> Result<TimeToSampleBox> {
 /// Parse a VPx Config Box.
 fn read_vpcc<T: Read>(src: &mut BMFFBox<T>) -> Result<VPxConfigBox> {
     let (version, _) = read_fullbox_extra(src)?;
-    if version != 0 {
+    if version != 0 && version != 1 {
```
alfredoyang: Yeah, I'll change to that way. Thanks!

mp4parse/src/tests.rs
alfredoyang: Fixed.

mp4parse/src/boxes.rs
alfredoyang: Removed.

PR #99, opened 2017-06-22
support ac3, ec3 atom for quicktime format

alfredoyang:
https://bugzilla.mozilla.org/show_bug.cgi?id=1374194

Support AC3/EC3 audio sample entry in QT [1].

[1] https://developer.apple.com/library/content/documentation/QuickTime/QTFF/QTFFChap3/qtff3.html#//apple_ref/doc/uid/TP40000939-CH205-SW1

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -267,6 +267,8 @@ pub enum AudioCodecSpecific {
     ES_Descriptor(ES_Descriptor),
     FLACSpecificBox(FLACSpecificBox),
     OpusSpecificBox(OpusSpecificBox),
+    AC3SpecificBox,
+    EC3SpecificBox,
     MP3,
 }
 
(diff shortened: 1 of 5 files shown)


mp4parse/src/lib.rs
```diff
@@ -267,6 +267,8 @@ pub enum AudioCodecSpecific {
     ES_Descriptor(ES_Descriptor),
     FLACSpecificBox(FLACSpecificBox),
     OpusSpecificBox(OpusSpecificBox),
+    AC3SpecificBox(Vec<u8>),
+    EC3SpecificBox(Vec<u8>),
```
kinetiknz: Unless we plan to support these codecs, it seems like a waste of time reading the box contents and storing them just to ignore them later.  Skipping over them would also simplify the parsing code slightly.  Thoughts?

alfredoyang: Yeah, I'll change it.

PR #100, opened 2017-06-27
Don't reject the negative timecodes.

alfredoyang:
https://bugzilla.mozilla.org/show_bug.cgi?id=1376190

The ```ctts``` box version 1 could produce a negative time. Since we treat all ```cts``` as version 1, the negative time is expectable.

[1] 14496-12 8.6.1.3.1


alfredoyang:
hmm... I don't get your point. The {start,end}_composition are actually i64.

```
pub struct mp4parse_indice {
    pub start_offset: u64,
    pub end_offset: u64,
    pub start_composition: i64,
    pub end_composition: i64,
    pub start_decode: i64,
    pub sync: bool,
}
```

Do you mean we should always validate the timecode?

kinetiknz:
Sorry for the confusion Alfredo, I must've been looking at an old tree before f30b9f0623d55f967c49b94c711c213ee79b9f02.

PR #107, opened 2017-07-18
use get() or get_mut() when access to a vector in case of out of vector boundary

alfredoyang:
Vector is a dynamic array. It could be out of boundary without checking. We need to use get() or get_mut() to get rid of this problem.

https://bugzilla.mozilla.org/show_bug.cgi?id=1359275


mp4parse_capi/src/lib.rs
```diff
@@ -989,11 +996,19 @@ fn create_sample_table(track: &Track, track_offset_time: i64) -> Option<Vec<mp4p
...
-            let peek_index = sort_table[i + 1];
-            let next_start_composition_time = sample_table[peek_index].start_composition;
-            let sample = &mut sample_table[current_index];
-            sample.end_composition = next_start_composition_time;
+            let (current_index, peek_index) = match (sort_table.get(i), sort_table.get(i + 1)) {
+                (Some(&current_index), Some(&peek_index)) => (current_index, peek_index),
+                _ => return None,
+            };
+
+            let next_start_composition_time = match sample_table.get(peek_index) {
```
kinetiknz: Do we need to use the fallible getters for peek_index and current_index?  Since they're extracted from sort_table which contains only valid indices into sample_table, it's impossible for these values to be out of range, unless I'm missing something.

alfredoyang: Yes, it won't out of range. This part looks over fix to this problem when I changed it. Just not sure if I'd always follow the coding guideline. I'll revert it back.

mp4parse_capi/src/lib.rs
```diff
@@ -884,7 +884,10 @@ fn create_sample_table(track: &Track, track_offset_time: i64) -> Option<Vec<mp4p
     for i in stsc_iter {
         let chunk_id = i.0 as usize;
         let sample_counts = i.1;
-        let mut cur_position: u64 = stco.offsets[chunk_id];
+        let mut cur_position: u64 = match stco.offsets.get(chunk_id) {
```
alfredoyang: No. :-)

PR #108, opened 2017-07-18
Correct descriptor length algorithm

alfredoyang:
The algorithm to count the descriptor length is incorrect before. (I should get an updated 14496-1 spec. )

[1] https://github.com/crosswalk-project/chromium-crosswalk/blob/master/media/formats/mp4/es_descriptor.cc#L14


alfredoyang:
No, I didn't find a file yet. I'll create a testcase manually.

PR #109, opened 2017-08-14
Reduce max table length in 32bit arch for OOM problem

alfredoyang:
This is from https://bugzilla.mozilla.org/show_bug.cgi?id=1389470. The VM space is 2G in the 32-bit environment which causes OOM when the table size is too large. The original length could use up to 800MB memory that causes OOM very easily.
So this patch reduces table size to one-day long video to avoid OOM.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -30,7 +30,12 @@ const BUF_SIZE_LIMIT: usize = 1024 * 1024;
 
 // Max table length. Calculating in worth case for one week long video, one
 // frame per table entry in 30 fps.
+#[cfg(target_pointer_width = "64")]
 const TABLE_SIZE_LIMIT: u32 = 30 * 60 * 60 * 24 * 7;
+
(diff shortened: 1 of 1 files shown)


kinetiknz:
(duplicating my comment from the BMO bug):

I think long term we want fallible allocation so we can handle OOM gracefully.  There's some discussion about this among the Rust language people (https://internals.rust-lang.org/t/notes-interviews-on-fallible-collection-allocations/5619 is the latest, I think), but I don't know how far away a solution is.  It's possible we could find a fallible Vec crate and switch to that (or implement our own), if this continues to be a problem with the adjusted limit.

For a short term fix, adjusting the limit based on the architecture seems reasonable.

kinetiknz:
Another option is to avoid these potentially large allocations entirely by changing the interface to a (streaming) iterator based one, rather than returning the complete table.  This would require changes on the Gecko side, and I haven't investigated how feasible it is, but it may be worth considering.

PR #112, opened 2017-08-25
check PSSH header length before calling Vec::with_capacity

alfredoyang:
https://bugzilla.mozilla.org/show_bug.cgi?id=1388618

Invalid ```pssh``` header size causes OOM.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -709,9 +709,8 @@ fn read_moov<T: Read>(f: &mut BMFFBox<T>, context: &mut MediaContext) -> Result<
 }
 
 fn read_pssh<T: Read>(src: &mut BMFFBox<T>) -> Result<ProtectionSystemSpecificHeaderBox> {
-    let mut box_content = Vec::with_capacity(src.head.size as usize);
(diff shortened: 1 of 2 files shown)


mp4parse/src/lib.rs
```diff
@@ -709,6 +709,9 @@ fn read_moov<T: Read>(f: &mut BMFFBox<T>, context: &mut MediaContext) -> Result<
 }
 
 fn read_pssh<T: Read>(src: &mut BMFFBox<T>) -> Result<ProtectionSystemSpecificHeaderBox> {
+    if src.head.size > BUF_SIZE_LIMIT as u64 {
+        return Err(Error::InvalidData("Invalid pssh data."));
+    }
     let mut box_content = Vec::with_capacity(src.head.size as usize);
     src.read_to_end(&mut box_content)?;
```
kinetiknz: This pattern should be using read_buf instead, which already handles this check.

alfredoyang: Yes, thanks for the suggestion.

kinetiknz:
Thanks!

PR #113, opened 2017-09-01
make sure sample_table length > 0 before using

alfredoyang:
https://bugzilla.mozilla.org/show_bug.cgi?id=1395244

```
SampleToChunkBox { samples: [
  SampleToChunk { first_chunk: 100663297, samples_per_chunk: 2, sample_description_index: 1 }, 
  SampleToChunk { first_chunk: 2, samples_per_chunk: 1, sample_description_index: 1 }] }
```

Invalid first_chunk causes an invalid range (100663297, 2) in ```stsc_iter```, so the sample_table doesn't have any item. That causes ```index out of bounds``` error in the bug.

I'll add testcase in gecko gtest.


kinetiknz:
Thanks!

PR #114, opened 2017-09-08
add fallible memory allocation for vector

alfredoyang:
Rust doesn't support fallible memory allocation which causes lots of OOM in https://bugzilla.mozilla.org/show_bug.cgi?id=1389470.

In https://bugzilla.mozilla.org/show_bug.cgi?id=1389009, it adds a crate for ```FallibleVec``` which calls the native memory allocation which doubles its capacity and returns a Result.

So in this PR, we call ```FallibleVec.try_push()``` for each mp4 table in case of OOM. Because the memory management could be different between pure rust and the gecko, so we add a function ```mp4parse_fallible_allocation()``` to enable it.

Frankly speaking, it more likes a temporary solution before Rust supports fallible allocation officially in https://github.com/rust-lang/rfcs/pull/2116.

### DIFF
--- a/fallible/Cargo.toml
+++ b/fallible/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "fallible"
+version = "0.0.1"
+authors = ["The Servo Project Developers"]
+license = "MPL-2.0"
+publish = false
+
+[lib]
+name = "fallible"
+path = "lib.rs"
+
+[dependencies]
+smallvec = "0.4"
--- a/mp4parse/Cargo.toml
+++ b/mp4parse/Cargo.toml
@@ -29,6 +29,7 @@ afl-plugin = { version = "0.1.1", optional = true }
 abort_on_panic = { version = "1.0.0", optional = true }
 bitreader = { version = "0.3.0" }
 num-traits = "0.1.37"
+fallible = { path = "../fallible" }
 
 [dev-dependencies]
 test-assembler = "0.1.2"
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -11,12 +11,14 @@ extern crate afl;
 extern crate byteorder;
 extern crate bitreader;
 extern crate num_traits;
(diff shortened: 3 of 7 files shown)


fallible/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
```
kinetiknz: Can you please add a repository key to the toml (or, if it's a direct copy from upstream, to avoid editing it) a README linking to the source repo?

mp4parse_capi/src/lib.rs
```diff
@@ -982,7 +991,9 @@ fn create_sample_table(track: &Track, track_offset_time: i64) -> Option<Vec<mp4p
         let mut sort_table = Vec::new();
         sort_table.reserve(sample_table.len());
```
kinetiknz: In this case the allocation happens here, not at the push call.

mp4parse/src/lib.rs
```diff
@@ -74,6 +91,8 @@ pub enum Error {
     NoMoov,
     /// Parse error caused by table size is over limitation.
     TableTooLarge,
+    /// Out of memory
+    OutOfMemory,
```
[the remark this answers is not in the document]
alfredoyang: Will fix it later.

mp4parse/Cargo.toml
```diff
@@ -29,6 +29,7 @@ afl-plugin = { version = "0.1.1", optional = true }
 abort_on_panic = { version = "1.0.0", optional = true }
 bitreader = { version = "0.3.0" }
 num-traits = "0.1.37"
+fallible = { path = "../fallible" }
```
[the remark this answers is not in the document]
alfredoyang: Will fix it later.

fallible/Cargo.toml
alfredoyang: It is from https://github.com/servo/servo/tree/master/components/fallible. I'll add a README.

PR #115, opened 2017-09-12
initialize the read buffer

alfredoyang:
In https://github.com/mozilla/mp4parse-rust/pull/114, it needs to allocate memory for ```read_buf()```; however, it just extends its capacity, its length stays zero. So we need to initialize it to correct its length. Otherwise, it returns error in
```
impl Read for mp4parse_io {
    fn read(&mut self, buf: &mut [u8]) {
        ...
        let rv = self.read.unwrap()(buf.as_mut_ptr(), buf.len(), self.userdata);
        if rv >= 0 {
            Ok(rv as usize)
```

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -83,10 +83,11 @@ pub fn vec_reserve<T>(vec: &mut Vec<T>, size: usize) -> std::result::Result<(), 
     Ok(())
 }
 
-fn allocate_read_buf(size: usize) -> std::result::Result<Vec<u8>, ()> {
+fn reserve_read_buf(size: usize) -> std::result::Result<Vec<u8>, ()> {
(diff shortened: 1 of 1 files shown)


alfredoyang:
Thanks!!! Fixing.

kinetiknz:
I meant calling set_len in the caller of allocate_read_buf (i.e. in read_buf itself), but I had only looked at the diff lines and not the complete code, so I missed that the infallible case returns vec[0; size] (i.e. the length and capacity are already set correctly), so your version is actually the correct one since the two variants now return consistent length Vecs... in which case allocate_read_buf is a better name than reserve_read_buf and I shouldn't have suggested the renaming!

alfredoyang:
That's ok, I'll change it back later.

PR #117, opened 2017-09-14
Don't (re)allocate if the new_cap is the same size as old_cap.

kinetiknz:
This fixes a bug where try_reserve(0) would call malloc(0) then construct a Vec via from_raw_parts with the valid resulting 0-sized allocation and a capacity of 0.  Rust's implementation of Vec (specifically in RawVec::dealloc_buffer and RawVec::current_layout) assumes that a 0-capacity RawVec never allocated so avoids calling dealloc in this case.

r? @alfredoyang


alfredoyang:
Thanks!

PR #127, opened 2017-11-17
Switch to cbindgen

nox:


rillian:
Thanks for the patches! This looks fine to me, but I want to verify it doesn't cause any problems on the gecko side before merging so we don't have to rewrite things twice.

kinetiknz:
mp4parse_parser is supposed to be opaque on the C side, so generating a structure with visible public fields as indicated by https://github.com/eqrion/cbindgen/issues/79#issuecomment-345683164 suggests something is wrong.  Maybe 3e5b1f0 should be reverted.

nox:
@kinetiknz I'll try but I believe it just moves the problem around.

nox:
@kinetiknz It doesn't help.

```c
struct Wrap;
typedef struct Wrap Wrap;

typedef struct {
  Wrap _0;
} Mp4parseParser;
```

kinetiknz:
Remove ` #[repr(C)] ` and it should generate an opaque struct, from a quick look at cbindgen's tests.

And the struct should have a comment explaining that we want Mp4parseParser  to be opaque, like the old code had on the Wrap struct.

nox:
Removing `#[repr(C)]` makes the C++ code compile now, thanks.

Remaining issue is linking problems, but I wonder if that's macOS-specific:

```
---- build_ffi_test stdout ----
	status: exit code: 2
--- stdout ---
rm -f test
rm -f *.a.out
rm -f *.o *.a
rm -f *.d
rustc -g --crate-type staticlib --crate-name mp4parse \
	  --emit dep-info,link=libmp4parse.a \
	   \
	  -L ../../target/debug/deps ../src/lib.rs \
	  2> libmp4parse.a.out || cat libmp4parse.a.out >&2
c++ -g -Wall -std=c++11 -I../include/ -c test.cc
c++ -g -Wall -std=c++11 -I../include/ -o test *.o libmp4parse.a -lSystem -lresolv -lc -lm

-- stderr ---
Undefined symbols for architecture x86_64:
  "mp4parse_log(bool)", referenced from:
      main::$_0::operator()(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&) const in test.o
  "mp4parse_new(Mp4parseIo const*)", referenced from:
      test_new_parser() in test.o
      test_arg_validation() in test.o
      test_arg_validation_with_parser() in test.o
      test_arg_validation_with_data(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in test.o
      read_file(char const*) in test.o
  "mp4parse_free(Mp4parseParser*)", referenced from:
      test_new_parser() in test.o
      test_arg_validation_with_parser() in test.o
      test_arg_validation_with_data(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in test.o
      read_file(char const*) in test.o
  "mp4parse_read(Mp4parseParser*)", referenced from:
      test_arg_validation() in test.o
      test_arg_validation_with_parser() in test.o
      test_arg_validation_with_data(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in test.o
      read_file(char const*) in test.o
  "mp4parse_get_track_info(Mp4parseParser*, unsigned int, Mp4parseTrackInfo*)", referenced from:
      test_arg_validation() in test.o
      test_arg_validation_with_parser() in test.o
      test_arg_validation_with_data(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in test.o
      read_file(char const*) in test.o
  "mp4parse_get_track_count(Mp4parseParser const*, unsigned int*)", referenced from:
      test_arg_validation_with_data(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in test.o
      read_file(char const*) in test.o
  "mp4parse_get_track_audio_info(Mp4parseParser*, unsigned int, Mp4parseTrackAudioInfo*)", referenced from:
      test_arg_validation() in test.o
      test_arg_validation_with_parser() in test.o
      test_arg_validation_with_data(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in test.o
  "mp4parse_get_track_video_info(Mp4parseParser*, unsigned int, Mp4parseTrackVideoInfo*)", referenced from:
      test_arg_validation() in test.o
      test_arg_validation_with_parser() in test.o
      test_arg_validation_with_data(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in test.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [test] Error 1
```

kinetiknz:
My guess is that the generated header doesn't have `extern "C" {` wrapping it, so the symbols `mp4parse_new` etc. are exported with C++ name mangling rather than C.

kinetiknz:
> My guess is that the generated header doesn't have extern "C" { wrapping it, so the symbols mp4parse_new etc. are exported with C++ name mangling rather than C.

```
diff --git a/mp4parse_capi/build.rs b/mp4parse_capi/build.rs
index 14fbff3..d24dae1 100644
--- a/mp4parse_capi/build.rs
+++ b/mp4parse_capi/build.rs
@@ -12,7 +12,15 @@ fn main() {
         c.header = Some(concat!(
             "// This Source Code Form is subject to the terms of the Mozilla Public\n",
             "// License, v. 2.0. If a copy of the MPL was not distributed with this\n",
-            "// file, You can obtain one at https://mozilla.org/MPL/2.0/.",
+            "// file, You can obtain one at https://mozilla.org/MPL/2.0/.\n\n",
+            "#ifdef __cplusplus\n",
+            "extern \"C\" {\n",
+            "#endif",
+        ).into());
+        c.trailer = Some(concat!(
+            "#ifdef __cplusplus\n",
+            "} /* extern \"C\" */\n",
+            "#endif",
         ).into());
         c.include_guard = Some("MP4PARSE_CAPI_H".to_owned());
         c.autogen_warning = Some(
```

It looks like cbindgen only generates `extern "C"` if the Config.language is Cxx, which makes some sort of sense.  But we want to generate a C compatible header that's also usable in C++, so it looks like we need to do the above to get what we want.

nox:
@kinetiknz Thanks again, builds should be green now.

kinetiknz:
Awesome, thanks!  Eyeballing the new `mp4parse.h` side-by-side with the old rusty-cheddar version, the minor differences look acceptable.

I prefer the old type names and it's a shame the ordering of the declarations in mp4parse_capi/src/lib.rs is lost in the new mp4parse.h.  Oh well, progress.

I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1419627 for the Gecko side changes.

rillian:
Thanks for the review and cleanup. Looks good to me now too!

PR #135, opened 2018-01-04
check descriptor length before parsing

alfredoyang:
There is an extra byte at the end of ```esds```, in the original design, it runs into io error when reading data. That's an invalid descriptor frankly speaking, but it is good if we could tolerate this kind of error and keep parsing instead of returning an error.

the whole esds:
  ```[3, 25, 0, 1, 0, 4, 19, 64, 21, 0, 0, 0, 0, 0, 0, 0, 0, 1, 119, 0, 5, 2, 18, 16, 6, 1, 2]```
es descritor within esds:
  ```[4, 19, 64, 21, 0, 0, 0, 0, 0, 0, 0, 0, 1, 119, 0, 5, 2, 18, 16, 6, 1, 2]```
dc descriptor within es:
  ```[5, 2, 18, 16, 6, 1]```

The final ```2``` is redundant data.

https://bugzilla.mozilla.org/show_bug.cgi?id=1426773

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -1376,7 +1376,8 @@ fn find_descriptor(data: &[u8], esds: &mut ES_Descriptor) -> Result<()> {
 
     let mut remains = data;
 
-    while !remains.is_empty() {
+    // Descriptor length should be more than 2 bytes.
+    while remains.len() > 2 {
         let des = &mut Cursor::new(remains);
         let tag = des.read_u8()?;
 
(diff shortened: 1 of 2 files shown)


alfredoyang:
hmm... it fails at test ```free_null_parser``` with nightly rustc. It looks like a bug in rustc nightly.

```
   1145	#[should_panic(expected = "assertion failed")]
   1146	fn free_null_parser() {
   1147	    unsafe {
-> 1148	        mp4parse_free(std::ptr::null_mut());
   1149	    }
   1150	}
   1151
```

kinetiknz:
Changes LGTM, thanks.

Looks like the build has been failing on nightly since the 27th Dec weekly cron job.  The last good build was the 20th (previous weekly cron job), so that gives us the window for the nightly regression.

PR #149, opened 2018-05-01
Parse uuid box types

hfiguiere:
This allow parsing 'uuid' boxes.

While this is generic to the ISO Base Media format, I implemented this to be able to parse Canon CR3 files (brand 'crx ') that use 'uuid' boxes.


kinetiknz:
Thanks!  This looks good.  Would you mind adding a new test to mp4parse/src/test.rs along the lines of read_box_header_short?  You'll probably need to extend make_box/make_full_box to take a UUID.

hfiguiere:
Added the test. Let me know if you prefer I squash the commits.

kinetiknz:
That's perfect, thank you.  I'll squash as part of the merge.

PR #151, opened 2018-05-11
Ignore truncated uuid boxes rather than raising an error.

kinetiknz:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1457359#c6

@jyavenard r?


jyavenard:
r+

PR #152, opened 2018-05-29
Misc improvements

hfiguiere:
- Recognize metadata tracks: they can matter.

- Move check_parser_state macro to boxes.rs so that it can be used by another module.

The purpose of these is to facilitate my fork that I use to parse Canon CR3 files. These changes are generic to the IsoBaseMedia.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -436,6 +436,7 @@ impl MediaContext {
 pub enum TrackType {
     Audio,
     Video,
+    Metadata,
     Unknown,
 }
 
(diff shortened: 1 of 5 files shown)


kinetiknz:
Thanks! This looks good.  I think it'd be better to move `check_parser_state` to a new macros.rs file, it seems a bit out of place in boxes.rs.

hfiguiere:
OK I'll do that (that was my initial approach)

hfiguiere:
Updated branch. The macro has been moved to `macros.rs`

kinetiknz:
Thanks!

PR #155, opened 2018-08-14
AV1 support.

tdaede:
This is functional, but needs a bit of cleanup.


kinetiknz:
Looks good so far, thanks.  read_av1c and the public_video_av1 test need to be fleshed out before merging, and the debug logging removed.  Let me know when you're ready for a full review.

tdaede:
OK, this has now been updated with a mp4 that matches the latest version of the spec, plus a test that matches all the fields.

kinetiknz:
Doc test is failing, I think due to a typo.  Once that's fixed are you ready for me to review and merge?

tdaede:
It should be ready to review, however the sample might change once more due to this issue on the mp4 spec dropping the fullbox header: https://github.com/AOMediaCodec/av1-isobmff/issues/71

kinetiknz:
LGTM, thanks.

Do you know if we'll need to support parsing `colr` boxes and the others related to HDR content, or will the decoder use the values from the sequence header/metadata OBUs?

Looking over this code reminds me that we should probably refactor `read_av1c`, `read_vpcc`, and probably any of the other bit manipulation code to use BitReader (introduced for ESDS parsing) sometime.

tdaede:
We probably won't, as unlike VP9, most of the color information is inside the AV1 bitstream.

PR #163, opened 2018-10-24
 Handle and expose multiple sample descriptions per track

SingingTree:
The mp4 spec allows for tracks to have multiple sample descriptions.
This is done to allow for different codings within a track. Cases of
this being done appear to be rare, and historically the parser was able
to assume a single description was used. However, instances of multiple
descriptions are appearring in encrypted content packaged by Google's
Shaka packager. To avoid compatibility issues, and as this behaviour is
specified, we should handle and expose multiple samples descriptions per
track.

Technically, this changes how we expose the contents of stsd boxes.
Previouslly we would map a combination of the contained entires to a
single `data` member of the associated track. With this change, a track
now exposes each of its sample descriptions. The data member on tracks
has been removed.

For consumers, if a single sample description is present on a track,
this can be handled in a similar fashion as with `track.data`. If
multiple sample descriptions exist, handling is more complex, and the
relevant sample description must be looked up for samples by the consumer.
However, this should be preferred to offering an incomplete view of
the metadata.

The c-api and tests are also updated to avoid bustages.

### DIFF
--- a/mp4parse/Cargo.toml
+++ b/mp4parse/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mp4parse"
-version = "0.10.1"
+version = "0.11.0"
 authors = [
   "Ralph Giles <giles@mozilla.com>",
   "Matthew Gregan <kinetik@flim.org>",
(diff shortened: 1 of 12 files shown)


kinetiknz:
Looks great, thank you!

PR #164, opened 2018-10-30
capi tests: use default() for structs and add back some codec checks.

SingingTree:
- There are a number of cases where structs that have default values are
being created with explicit values in the capi tests. We can just use
default(), it makes the tests more concise, and avoids confusion where
readers may think we're intentionally using explicit values for
programatic reasons.
- Recent changes have moved codec information from track structures to
sample description structures -- as a single track may have multiple
different sample descriptions with their own codecs. In doing so, I had
removed some codec checks from our tests. Restore some of these checks
in test_fragment.rs, so that we have coverage, and as that test inspects
at the sample description level.


kinetiknz:
Thanks!

PR #166, opened 2018-11-08
Handle schm box

SingingTree:
The SchemeType box stores a four character code representing the
encryption scheme used for a track. It also stores a version number for
the scheme used. Historically the scheme could be determined by other
metadata, without explicitly reading the FourCC. This appears to still be
the case, however, parsing this information is useful because:
- It's easier to read the encryption scheme from the scheme type box
that deduce it from other values.
- We can use the explicit scheme type to checksum the other encryption
values and make sure they're sane for a given type.
- If I'm wrong about being able to deduce all schemes from other
metadata, we may require the scheme type to handle certain encrypted
media.

The scheme type is read and exposed in both the rust and C APIs. The
scheme version is currently not exposed for C, as it doesn't appear to
have an impact to how MP4s should be handled in the current spec.

Add tests to verify scheme parsing for encrypted and non encrypted
cases.

Bump version numbers given the API change.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -425,6 +425,12 @@ pub struct ProtectionSystemSpecificHeaderBox {
 }
 
 #[derive(Debug, Default, Clone)]
+pub struct SchemeTypeBox {
+    pub scheme_type: FourCC,
+    pub scheme_version: u32,
+}
+
+#[derive(Debug, Default, Clone)]
 pub struct TrackEncryptionBox {
     pub is_encrypted: u8,
     pub iv_size: u8,
@@ -439,6 +445,7 @@ pub struct TrackEncryptionBox {
 #[derive(Debug, Default, Clone)]
 pub struct ProtectionSchemeInfoBox {
     pub code_name: String,
+    pub scheme_type: Option<SchemeTypeBox>,
     pub tenc: Option<TrackEncryptionBox>,
 }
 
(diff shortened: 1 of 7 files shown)


mp4parse/src/lib.rs
```diff
@@ -439,6 +445,7 @@ pub struct TrackEncryptionBox {
 #[derive(Debug, Default, Clone)]
 pub struct ProtectionSchemeInfoBox {
     pub code_name: String,
+    pub scheme_type: Option<SchemeTypeBox>,
```
SingingTree: Wasn't sure if there's a convention around naming such members. I considered naming this similar to `tenc`, but prefer the more verbose naming for those who are not familiar with the FourCC. Am happy to change to `schm` if it's more fitting.

mp4parse/src/lib.rs
```diff
@@ -2203,6 +2213,19 @@ fn read_frma<T: Read>(src: &mut BMFFBox<T>) -> Result<String> {
     String::from_utf8(code_name).map_err(From::from)
 }
 
+fn read_schm<T: Read>(src: &mut BMFFBox<T>) -> Result<SchemeTypeBox> {
+    // Flags can be used to signal presence of URI in the box, but we don't
+    // use the URI so don't bother storing the flags.
+    let (_, _) = read_fullbox_extra(src)?;
+    let scheme_type =  FourCC::from(be_u32(src)?);
+    let scheme_version = be_u32(src)?;
+    // Null terminated scheme URI may follow, but we don't use it right now.
```
kinetiknz: Even if we don't need the scheme_uri, we need to skip the rest of the box contents or we'll trigger a parser desync if it's present.  `skip_box_remain` will take care of that.

SingingTree:
Looks like Travis may have changed OSX pre-installed config. Will look into resolving that in another PR then then waking this one.

PR #167, opened 2018-11-11
Remove oclint work around for travis now that it's not needed.

SingingTree:
Per [this](https://changelog.travis-ci.com/oclint-is-removed-from-mac-builds-79270) we no longer need the work around for travis, and indeed need it removed to avoid failures.

Edit: should it be useful, see also https://github.com/travis-ci/travis-ci/issues/8826 for historical context.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,6 @@ os:
   - linux
   - osx
   - windows
-
-before_install:
-  # Work around https://github.com/travis-ci/travis-ci/issues/8826, when that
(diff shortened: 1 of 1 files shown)


SingingTree:
Looking like a networking error trying to install gcc5 for OSX 🙁 @kinetiknz, could you kick off another CI run when you have a moment to see if this is a transient issue? Otherwise I'll have to do something else to work around this, or failing that, will disable the OSX runs.

kinetiknz:
Thanks for investigating and fixing!

PR #173, opened 2019-02-19
Zero initialize Vec returned from mp4parse_fallible branch of allocate_read_buf.

### LINKED ISSUE #172: read_mp4 can expose uninitialised data to third party code (unsound)
[`read_mp4`](https://github.com/mozilla/mp4parse-rust/blob/26e614bbcb4d7322dc2e4b7e15391bdb30b9f7be/mp4parse/src/lib.rs#L700) calls [`read_moov`](https://github.com/mozilla/mp4parse-rust/blob/26e614bbcb4d7322dc2e4b7e15391bdb30b9f7be/mp4parse/src/lib.rs#L729) which calls [`read_pssh`](https://github.com/mozilla/mp4parse-rust/blob/26e614bbcb4d7322dc2e4b7e15391bdb30b9f7be/mp4parse/src/lib.rs#L783) which calls [`read_buf`](https://github.com/mozilla/mp4parse-rust/blob/26e614bbcb4d7322dc2e4b7e15391bdb30b9f7be/mp4parse/src/lib.rs#L796) which calls [`allocate_read_buf`](https://github.com/mozilla/mp4parse-rust/blob/26e614bbcb4d7322dc2e4b7e15391bdb30b9f7be/mp4parse/src/lib.rs#L2320) and passes its result to [`Read::read`](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read).

https://github.com/mozilla/mp4parse-rust/blob/26e614bbcb4d7322dc2e4b7e15391bdb30b9f7be/mp4parse/src/lib.rs#L2320-L2321

But the trait `Read` is not unsafe, and it is never guaranteed to limit itself to *writing* to its single argument `buf`, and the result of `allocate_read_buf` is a vector of uninitialised bytes (that function should be unsafe, btw).

There were discussions about introducing a new unsafe trait in libstd to signal that a `Read` implementation doesn't read in the writer it's supposed to write to, I don't know what became of it but I just found out about https://github.com/rust-lang/rust/issues/42788 which seems related.

kinetiknz:
Zero initialize Vec returned from mp4parse_fallible branch of allocate_read_buf.

Also delete vec_reserve, which is completely unused.

Closes #172.

r? @SingingTree please

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -42,8 +42,8 @@ const BUF_SIZE_LIMIT: usize = 1024 * 1024;
 // frame per table entry in 30 fps.
 const TABLE_SIZE_LIMIT: u32 = 30 * 60 * 60 * 24 * 7;
 
-// TODO: vec_push() and vec_reserve() needs to be replaced when Rust supports
-// fallible memory allocation in raw_vec.
+// TODO: vec_push() needs to be replaced when Rust supports fallible memory
+// allocation in raw_vec.
 #[allow(unreachable_code)]
 pub fn vec_push<T>(vec: &mut Vec<T>, val: T) -> std::result::Result<(), ()> {
     #[cfg(feature = "mp4parse_fallible")]
@@ -56,23 +56,12 @@ pub fn vec_push<T>(vec: &mut Vec<T>, val: T) -> std::result::Result<(), ()> {
 }
 
 #[allow(unreachable_code)]
-pub fn vec_reserve<T>(vec: &mut Vec<T>, size: usize) -> std::result::Result<(), ()> {
-    #[cfg(feature = "mp4parse_fallible")]
-    {
-        return FallibleVec::try_reserve(vec, size);
-    }
-
-    vec.reserve(size);
-    Ok(())
-}
-
-#[allow(unreachable_code)]
 fn allocate_read_buf(size: usize) -> std::result::Result<Vec<u8>, ()> {
     #[cfg(feature = "mp4parse_fallible")]
     {
         let mut buf: Vec<u8> = Vec::new();
         FallibleVec::try_reserve(&mut buf, size)?;
-        unsafe { buf.set_len(size); }
+        buf.extend(std::iter::repeat(0).take(size));
         return Ok(buf);
     }
 


mp4parse/src/lib.rs
```diff
@@ -55,24 +55,16 @@ pub fn vec_push<T>(vec: &mut Vec<T>, val: T) -> std::result::Result<(), ()> {
...
 fn allocate_read_buf(size: usize) -> std::result::Result<Vec<u8>, ()> {
     #[cfg(feature = "mp4parse_fallible")]
     {
         let mut buf: Vec<u8> = Vec::new();
         FallibleVec::try_reserve(&mut buf, size)?;
-        unsafe { buf.set_len(size); }
+        unsafe {
+            buf.set_len(size);
+            std::ptr::write_bytes(buf.as_mut_ptr(), 0, buf.len() * std::mem::size_of::<u8>());
+        }
```
nox: You can do `buf.extend(std::iter::repeat(0).take(size))` to avoid unsafe code. A loop on  `0..size` with `Vec::push` calls would probably also compile down to the same thing.

SingingTree: Possibly done the way it is to avoid needing to reserve more space? Let me know if I'm mistaken, but if we `try_reserve(buf, n)`, `buf.set_len(n)`, then zero the vec with a ptr, we don't trigger further reserves (`set_len` guarantees it won't modify buffers). If we extend or push we get will get an extra allocation when we push at len == cap.

mp4parse/src/lib.rs
```diff
@@ -55,24 +55,16 @@ pub fn vec_push<T>(vec: &mut Vec<T>, val: T) -> std::result::Result<(), ()> {
...
-}
-
 #[allow(unreachable_code)]
 fn allocate_read_buf(size: usize) -> std::result::Result<Vec<u8>, ()> {
     #[cfg(feature = "mp4parse_fallible")]
     {
         let mut buf: Vec<u8> = Vec::new();
         FallibleVec::try_reserve(&mut buf, size)?;
-        unsafe { buf.set_len(size); }
+        unsafe {
```
SingingTree: Following on from my comment below. If I'm not incorrect about the reasoning, a comment here as to that we're doing this specifically to avoid another reserve would be useful.

mp4parse/src/lib.rs
nox: This is incorrect, the `Vec` API won’t reserve additional space in the case the iteration fits, and it does here.

nox: Note that [`Vec::push`](https://github.com/rust-lang/rust/blob/74e35d270067afff72034312065c48e6d8cfba67/src/liballoc/vec.rs#L1019-L1030) itself will never reallocate as long as it has space, too. In general, don't use unsafe code if you don't need to.

SingingTree: Thanks for the correction. I was mistaken as to when the new reserves take place (I was under the impression it was at the time len hit cap).

kinetiknz:
Thanks for the feedback from both of you.  Updated to use extend.

PR #72, opened 2017-02-28
Build with --release on travis.

rillian:
Add another dimension to the build matrix testing `cargo build --release` as well as the dev profile.

PR #181, opened 2019-07-05
Build with --release on Travis.

kinetiknz:
Fresh attempt at #72.


kinetiknz:
`--release` tests are built with Cargo's `bench` profile, so `overflow-checks` needs to be enabled for both `release` and `bench` profiles.

PR #182, opened 2019-07-09
Implement parsing for iTunes metadata atoms

### LINKED ISSUE #150: Implement `meta` parsing
Apple's [QuickTime File Format](https://developer.apple.com/library/content/documentation/QuickTime/QTFF/Metadata/Metadata.html#//apple_ref/doc/uid/TP40000939-CH1-SW1) is the closest to a standard to storing metadata in MP4 Boxes, and should be sub boxes of `trak`, `moov` or `mdia`

chyyran:
Fixes #150. 

Metadata in .MP4 files can be stored in either iTunes-style, or 3GPP-style formats. The most commonly found today is the iTunes-style format, which this Pull Request implements parsing for. This is useful for mp4parse to serve as a Rust library that can read these tags off MP4 audio and video files, and can be used as a foundation for a pure-Rust TagLib-like library, along with other metadata-parsing libraries available on Cargo.

iTunes metadata is stored within a Metadata `meta` atom within a root-level User data `udta` atom. While this `meta` atom may contain other atoms such as `keys`, we are mostly concerned with the `ilst` List atom, which holds a list of various named atoms that represent metadata. 

Each metadata box or entry within `ilst` has a unique key and is parsed as a full box, within which a `data` atom that contains the actual data, in either a string or binary `u8` format, is contained. Each metadata entry may contain only one `data` atom, with the exception of the cover art album `covr`, which may have multiple `data` atoms that each contain JPEG or PNG data. With the exception of `©lyr` and `ldes`, string metadata tags are limited to 256 characters in length, while binary metadata often have idiosyncratic formats that must be special-cased.

Parsing for all if not most common tags have been implemented, with the exception of the iTunes specific iTMS `xxID` entries, due to lack of documentation on these tags.

The `udta` and `meta` boxes are not exposed in the C API, since Firefox does not have a current need to use these atoms. They are currently only exposed in the Rust API.

This [public domain picture of a black hole](https://commons.wikimedia.org/wiki/File:Black_hole.jpg) is embedded within `metadata.mp4` and `metadata_gnre.mp4` for testing the `covr` atom. The test files themselves are just copies of `minimal.mp4` with a bunch of tags added to them.

A huge thanks to the AtomicParsley project for documentation on the format of individual tag boxes.


mp4parse/src/lib.rs
```diff
@@ -439,14 +439,188 @@ pub struct ProtectionSchemeInfoBox {
     pub tenc: Option<TrackEncryptionBox>,
 }
 
+/// Represents a userdata box 'udta'.
+/// Currently, only the metadata atom 'meta'
+/// is parsed.
+#[derive(Debug, Default, Clone)]
+pub struct UserdataBox {
+    pub meta: Option<MetadataBox>
```
chyyran: While calling the `udta` box `userdata` in the context is probably a reasonable decision, I'm not sure if the field for the nested `meta` box should be named `metadata`, since "metadata" can be ambiguous between track metadata, and metadata tags relating to the media the file or stream contains. Since the name of the box itself is `meta`, and it's a relatively self-explanatory FourCC, keeping it as `meta` makes sense, but naming it `metadata` would also be more consistent. 

If you have an opinion either way, would be happy to change it.

chyyran:
@kinetiknz could I get a review please?

mp4parse/src/lib.rs
```diff
@@ -2286,6 +2465,184 @@ fn read_schm<T: Read>(src: &mut BMFFBox<T>) -> Result<SchemeTypeBox> {
...
+            BoxType::CoverArtEntry => meta.cover_art = read_multiple_u8_data(&mut b).ok(),
+            _ => skip_box_content(&mut b)?,
+
+        };
+        check_parser_state!(b.content);
+    }
+    Ok(())
+}
+
+fn read_bool_data<T: Read>(src: &mut BMFFBox<T>) -> Result<Option<bool>> {
```
kinetiknz: Rename these new read_* additions to read_ilst_* or move them inside read_ilst to make it clear they're ILST-specific.

mp4parse/src/lib.rs
```diff
@@ -2286,6 +2465,184 @@ fn read_schm<T: Read>(src: &mut BMFFBox<T>) -> Result<SchemeTypeBox> {
...
+        };
+        check_parser_state!(b.content);
+    }
+    Ok(data)
+}
+
+fn read_data<T: Read>(src: &mut BMFFBox<T>) -> Result<Vec<u8>> {
+    // Skip past the padding bytes
+    skip(&mut src.content, src.head.offset as usize)?;
+    let size =  src.content.limit() as usize;
```
kinetiknz: Remove extra space after =.

mp4parse/src/lib.rs
```diff
@@ -2286,6 +2465,184 @@ fn read_schm<T: Read>(src: &mut BMFFBox<T>) -> Result<SchemeTypeBox> {
...
+fn read_string_data<T: Read>(src: &mut BMFFBox<T>) -> Result<Option<String>> {
+    read_u8_data(src)?
+        .map_or(Ok(None),
+                |d| String::from_utf8(d)
+                    .map_err(From::from)
+                    .map(Some)
+        )
+}
+
+
```
kinetiknz: Drop double newlines here and below.

mp4parse/src/lib.rs
```diff
@@ -2286,6 +2465,184 @@ fn read_schm<T: Read>(src: &mut BMFFBox<T>) -> Result<SchemeTypeBox> {
...
+            }
+            _ => skip_box_content(&mut b)?,
+        };
+        check_parser_state!(b.content);
+    }
+    Ok(data)
+}
+
+
+fn read_multiple_u8_data<T: Read>(src: &mut BMFFBox<T>) -> Result<Vec<Vec<u8>>> {
```
kinetiknz: It'd be nice if we could make most of read_multiple_u8_data and read_u8_data common.

mp4parse/src/lib.rs
```diff
@@ -773,6 +947,11 @@ fn read_moov<T: Read>(f: &mut BMFFBox<T>, context: &mut MediaContext) -> Result<
                 debug!("{:?}", pssh);
                 vec_push(&mut context.psshs, pssh)?;
             }
+            BoxType::UserdataBox => {
+                let udta = read_udta(&mut b)?;
```
kinetiknz: My primary concern with these additions is that the main user of mp4parse-rust (Gecko) is exposed to new forms of parse failure if invalid udta (or children) are encountered.  We don't want to reject MP4s we previously accepted due to content Gecko isn't interested in.

I think a simple solution to this is to make MediaContext::userdata a `Result<Option<Userdata>>` instead.  That does make the MediaContext API slightly odd, though.

chyyran: Instead of `Result<Option<UserdataBox>>`, I elected to make it `Option<Result<UserdataBox>>` instead, since `Result<Option<_>>` doesn't implement `Default`. 

I considered just ignoring the error and returning `None` if udta parsing fails, so as to not introduce any weirdness into the MediaContext API, but that seems to me that it would conflate the lack of an udta atom with a corrupt udta atom.

kinetiknz:
Thanks, looks good!

Issue #176, opened 2019-04-03
Update VPxConfigBox field names to match v1.0 of the vpcC spec


kinetiknz:
@jyavenard noticed the field names of VPxConfigBox don't match the ([current v1.0](https://www.webmproject.org/vp9/mp4/)) vpcC spec.

This is because the names are based on v0.0 of the spec (from Netflix/vp9-dash).  We should rename the VPxConfigBox fields to match v1.0, since v0.0 was experimental(ish) and is probably dead now.  The existing v0.0 support will be retained by mapping the v0.0 fields to v1.0.

PR #184, opened 2019-10-25
Update vpx config box names

SingingTree:
Seeks to fix https://github.com/mozilla/mp4parse-rust/issues/176 and add some more documentation along the way.

Let me know if some other formatting is preferred for the multilines that were broken. I based this on rustfmt's defaults.


kinetiknz:
LGTM, thanks!

PR #186, opened 2019-10-30
Add clippy config to travis CI.

SingingTree:
This should have travis run clippy as well as fail if any warnings are
found (not just in clippy, but elsewhere too).

I *think* this should work, but given the matrix of configs it's possible something breaks, so suggest we keep an eye on travis for a bit if we merge this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,11 @@ env:
     - RELEASE=
     - RELEASE=--release
 
+# Make sure clippy is installed, if it fails on nightly fall back to installing from repo.
+before_script:
+   - rustup component add clippy || cargo install --git https://github.com/rust-lang/rust-clippy/ --force clippy
+
 script:
+  - cargo clippy --all --all-targets $RELEASE -- -D warnings
   - cargo test --all --verbose $RELEASE
   - cargo doc --package mp4parse_capi


SingingTree:
Looks like something is already unhappy. I'll iterate some more on this and update the PR once I've clobbered the issues.

SingingTree:
Looks like I need to fix:
- Some of the tests don't play nicely if you check before running the tests. I.e. if you `cargo check` or `cargo clippy` then run `cargo test --all` the tests fail. Not sure why yet, but it's a problem that isn't unique to these changes.
- There look to be some nightly clippy lints that I missed.

SingingTree:
#187 looks to fix the nightly lint issue. Now to try and fix the issue of `cargo check` before `cargo test` breaking the tests.

SingingTree:
#197 is the issue blocking this.

baumanj:
Could you do a similar workaround to https://github.com/mozilla/mp4parse-rust/pull/193/commits/6c13760df485e213bb7058fb1751b9c4657ec21d and move the clippy run *after* `cargo test`?

SingingTree:
> 
> 
> Could you do a similar workaround to [6c13760](https://github.com/mozilla/mp4parse-rust/commit/6c13760df485e213bb7058fb1751b9c4657ec21d) and move the clippy run _after_ `cargo test`?

Indeed, and it's probably what I should do since I haven't come up with a better solution. That, or doing an extra clean felt like hacks, but probably better than letting this PR rot. I'd been hacking on this in my spare time rather than as a mozilla task, so I'll look for some spare time to get this across the line.

baumanj:
Can this be reexamined in light of https://github.com/mozilla/mp4parse-rust/pull/281 ?

SingingTree:
Superseded by #282.

PR #187, opened 2019-11-01
Fix clippy lint warnings from nightly version of clippy.

SingingTree:
This fixes clippy lints that show up in nightly that I'd overlooked in https://github.com/mozilla/mp4parse-rust/pull/185

This is largely adding safety sections to the unsafe pub functions on the capi. These are largely similar in that the functions are dereferencing raw pointers, so the callers need to provide sane pointers.

There is one additional case of using `map` instead of `and_then` on an optional to avoid rewrapping a `Some`.


kinetiknz:
Thanks!

PR #192, opened 2020-02-05
Replace casts with infallible or error-checked conversions

baumanj:
Add `to_u64` and `to_usize` trait functions for infallible conversions
which can be checked at compile time to be platform safe.

Additionally, convert some arithmetic to checked versions so we can
error rather than panic or give invalid results on overflow.

Remove some unnecessary trait bounds on struct definitions.

Update `skip` to avoid allocating a buffer.


baumanj:
Fixed the commit message.

I think I'll need you to merge for me. I don't think I've been added to the organization yet, have I?

kinetiknz:
> I think I'll need you to merge for me. I don't think I've been added to the organization yet, have I?

Not yet - I only filed the bug request today.

Issue #195, opened 2020-02-28
Check for valid context on C API getters


baumanj:
Functions like [`mp4parse_get_indice_table`](https://github.com/mozilla/mp4parse-rust/blob/bbb4a1217d2826e95b0a843e73c9e1a2faa23953/mp4parse_capi/src/lib.rs#L900) access data through a context object within the `parser` parameter. To help reduce the risk of this unsafe code, the context object could keep some state to indicate whether a a valid parse operation preceded the getter call rather than assuming good data from the C side.

If the C caller passed in uninitialized memory, there's no guarantee that it wouldn't match the field that marked a context as valid, but it would be much less likely.

kinetiknz:
That's what Mp4parseParser::poisoned is intended for.  `mp4parse_get_indice_table` already tests it.  Are there cases you're thinking of that aren't covered by this?

That mechanism doesn't handle the C API receiving an invalid context pointer. We'd need to track valid pointers/handles returned from our API and reject anything else being passed in.  Not sure that's worth it, though.

baumanj:
The case I'm imagining is different from `poisoned`, which gets set [when a context _becomes_ invalid](https://github.com/mozilla/mp4parse-rust/blob/63d4744905b0a918ed76461d0de48f8188325f6b/mp4parse_capi/src/lib.rs#L423). I'm thinking of the state when a context exists because [`mp4parse_new`](https://github.com/mozilla/mp4parse-rust/blob/63d4744905b0a918ed76461d0de48f8188325f6b/mp4parse_capi/src/lib.rs#L367) has been called, but that context isn't really valid because [`mp4_parse_read`](https://github.com/mozilla/mp4parse-rust/blob/63d4744905b0a918ed76461d0de48f8188325f6b/mp4parse_capi/src/lib.rs#L409) hasn't been called.

It would be possible to repurpose `poisoned` to handle this case by setting it to `true` to begin with and only setting it to false when `mp4_parse_read` returned successfully. Though in that case, it might make more sense to use a different name like `parsed_ok` or something. Or, perhaps it would make more sense to have `mp4parse_new` do the work that `mp4_parse_read` does as well (and remove `mp4_parse_read`) so that the C side would never have access to a context that didn't represent a validly parsed file.

djg:
[...]

PR #200, opened 2020-03-06
Merge mp4parse[_avif]_read into mp4parse[_avif]_new

baumanj:
See https://github.com/mozilla/mp4parse-rust/issues/195 and individual commit messages

### DIFF
--- a/mp4parse_capi/examples/test.cc
+++ b/mp4parse_capi/examples/test.cc
@@ -16,12 +16,6 @@
 
 #include "mp4parse.h"
 
-intptr_t abort_read(uint8_t *buffer, uintptr_t size, void *userdata)
-{
-  // This shouldn't be called when allocating a parser.
-  abort();
-}
-
 intptr_t error_read(uint8_t *buffer, uintptr_t size, void *userdata)
 {
   return -1;
@@ -50,12 +44,6 @@ void test_arg_validation()
   assert(parser == nullptr);
 
   Mp4parseIo io = { nullptr, nullptr };
-  rv = MP4PARSE_STATUS_INVALID;
-  parser = mp4parse_new(&io, &rv);
-  assert(rv == MP4PARSE_STATUS_BAD_ARG);
-  assert(parser == nullptr);
-
-  io = { abort_read, nullptr };
   rv = MP4PARSE_STATUS_INVALID;
   parser = mp4parse_new(&io, &rv);
   assert(rv == MP4PARSE_STATUS_BAD_ARG);
--- a/mp4parse_capi/examples/test.cc
+++ b/mp4parse_capi/examples/test.cc
@@ -39,36 +39,34 @@ intptr_t io_read(uint8_t *buffer, uintptr_t size, void *userdata)
   return r;
 }
 
-void test_new_parser()
-{
-  int dummy_value = 42;
-  Mp4parseIo io = { abort_read, &dummy_value };
-  Mp4parseParser *parser = mp4parse_new(&io);
-  assert(parser != nullptr);
-  mp4parse_free(parser);
-  assert(dummy_value == 42);
-}
-
 void test_arg_validation()
 {
-  Mp4parseParser *parser = mp4parse_new(nullptr);
+  Mp4parseParser *parser = mp4parse_new(nullptr, nullptr);
+  assert(parser == nullptr);
+
+  Mp4parseStatus rv = MP4PARSE_STATUS_INVALID;
+  parser = mp4parse_new(nullptr, &rv);
+  assert(rv == MP4PARSE_STATUS_BAD_ARG);
   assert(parser == nullptr);
 
   Mp4parseIo io = { nullptr, nullptr };
-  parser = mp4parse_new(&io);
+  rv = MP4PARSE_STATUS_INVALID;
+  parser = mp4parse_new(&io, &rv);
+  assert(rv == MP4PARSE_STATUS_BAD_ARG);
   assert(parser == nullptr);
 
   io = { abort_read, nullptr };
-  parser = mp4parse_new(&io);
+  rv = MP4PARSE_STATUS_INVALID;
+  parser = mp4parse_new(&io, &rv);
+  assert(rv == MP4PARSE_STATUS_BAD_ARG);
   assert(parser == nullptr);
 
   int dummy_value = 42;
   io = { nullptr, &dummy_value };
-  parser = mp4parse_new(&io);
-  assert(parser == nullptr);
-
-  int32_t rv = mp4parse_read(nullptr);
+  rv = MP4PARSE_STATUS_INVALID;
+  parser = mp4parse_new(&io, &rv);
   assert(rv == MP4PARSE_STATUS_BAD_ARG);
+  assert(parser == nullptr);
 
   Mp4parseTrackInfo info;
   rv = mp4parse_get_track_info(nullptr, 0, &info);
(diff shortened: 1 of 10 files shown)


mp4parse_capi/src/lib.rs
```diff
@@ -385,61 +391,99 @@ impl Read for Mp4parseIo {
...
+}
+
+fn mp4parse_new_common_safe<T: Read, P: ContextParser>(io: &mut T) -> Result<*mut P, Mp4parseStatus> {
+    let mut context = P::Context::default();
+
+    P::read(io, &mut context)
+        .map(|_| P::with_context(context))
+        .map(Box::new)
+        .map(Box::into_raw)
+        .map_err(Mp4parseStatus::from)
```
baumanj: I realize this functional style perhaps clashes with the approach in the rest of this library. I thought it was cleaner than the alternative, but I'm happy to change if y'all prefer the equivalent:
```rust
match P::read(io, &mut context) {
    Ok(_) => {
        let parser = Box::new(P::with_context(context));
        Ok(Box::into_raw(parser))
    }
    Err(e) => Err(e.into()),
}
```

baumanj:
In case you're curious, it appears the required changes to Firefox non-test code should be minimal:
```diff
diff --git a/dom/media/mp4/MP4Metadata.cpp b/dom/media/mp4/MP4Metadata.cpp
--- a/dom/media/mp4/MP4Metadata.cpp
+++ b/dom/media/mp4/MP4Metadata.cpp
@@ -90,17 +90,19 @@ static intptr_t read_source(uint8_t* buf
 MP4Metadata::MP4Metadata(ByteStream* aSource)
     : mSource(aSource), mSourceAdaptor(aSource) {
   DDLINKCHILD("source", aSource);
-
-  Mp4parseIo io = {read_source, &mSourceAdaptor};
-  mParser.reset(mp4parse_new(&io));
-  MOZ_ASSERT(mParser);
 }
 
 MP4Metadata::~MP4Metadata() {}
 
 nsresult MP4Metadata::Parse() {
-  Mp4parseStatus rv = mp4parse_read(mParser.get());
-  if (rv != MP4PARSE_STATUS_OK) {
+  Mp4parseIo io = {read_source, &mSourceAdaptor};
+  Mp4parseStatus rv = MP4PARSE_rv_INVALID;
+  Mp4parseParser *parser = mp4parse_new(&io, &rv);
+  if (rv == MP4PARSE_rv_OK && parser) {
+    mParser.reset(parser);
+    MOZ_ASSERT(mParser);
+  } else {
+    MOZ_ASSERT(!mParser);
     MOZ_LOG(gMP4MetadataLog, LogLevel::Debug,
             ("Parse failed, return code %d\n", rv));
     return rv == MP4PARSE_STATUS_OOM ? NS_ERROR_OUT_OF_MEMORY
```

mp4parse_capi/src/lib.rs
```diff
@@ -300,6 +300,17 @@ pub struct Mp4parseParser {
     video_track_sample_descriptions: HashMap<u32, Vec<Mp4parseTrackVideoSampleInfo>>,
 }
 
+/// A unified interface for the parsers which have different contexts, but
+/// share the same pattern of construction. This allows unification of
+/// argument validation from C and minimizes the surface of unsafe code.
+trait ContextParser where Self: Sized {
+    type Context: Default;
+
+    fn with_context(context: Self::Context) -> Self;
```
kinetiknz: new_with_context or from_context?

mp4parse_capi/src/lib.rs
```diff
@@ -385,61 +391,95 @@ impl Read for Mp4parseIo {
...
-        poisoned: false,
-        opus_header: HashMap::new(),
-        pssh_data: Vec::new(),
-        sample_table: HashMap::new(),
-        audio_track_sample_descriptions: HashMap::new(),
-        video_track_sample_descriptions: HashMap::new(),
-    });
-
-    Box::into_raw(parser)
+pub unsafe extern fn mp4parse_new(io: *const Mp4parseIo, status_out: *mut Mp4parseStatus) -> *mut Mp4parseParser {
```
kinetiknz: Any reason not to return Mp4parseStatus and make Mp4parseParser the out parameter?

mp4parse_capi/examples/test.cc
```diff
@@ -39,36 +39,34 @@ intptr_t io_read(uint8_t *buffer, uintptr_t size, void *userdata)
...
-  parser = mp4parse_new(&io);
+  rv = MP4PARSE_STATUS_INVALID;
+  parser = mp4parse_new(&io, &rv);
+  assert(rv == MP4PARSE_STATUS_BAD_ARG);
   assert(parser == nullptr);
 
   io = { abort_read, nullptr };
-  parser = mp4parse_new(&io);
+  rv = MP4PARSE_STATUS_INVALID;
+  parser = mp4parse_new(&io, &rv);
```
kinetiknz: `abort_read` only existed to verify that initializing a parser didn't call into the IO context.  That's no longer true, so it's obsolete and can be removed now.

mp4parse_capi/src/lib.rs
```diff
@@ -1587,34 +1535,42 @@ fn get_track_count_null_parser() {
...
-        let parser = mp4parse_new(&io);
+        let mut rv = Mp4parseStatus::Invalid;
+        let parser = mp4parse_new(&io, &mut rv);
+        assert_eq!(rv, Mp4parseStatus::BadArg);
         assert!(parser.is_null());
 
         let io = Mp4parseIo { read: Some(panic_read),
                                userdata: null_mut };
-        let parser = mp4parse_new(&io);
+        let mut rv = Mp4parseStatus::Invalid;
```
kinetiknz: panic_read is now obsolete per abort_read

mp4parse_capi/src/lib.rs
baumanj: That was what I tried first. With this signature:
```rust
pub unsafe extern fn mp4parse_new(io: *const Mp4parseIo, parser_out: *mut *const Mp4parseParser) -> Mp4parseStatus {
```
It led to some very awkward and error-prone code on the test side. My first attempt (that compiled) looked like this:
```rust
let mut parser: *mut Mp4parseParser = std::ptr::null_mut();
let rv = mp4parse_new(&io, &mut (parser as *const _));
```
But is subtly wrong (can you see the bug?). The "correct" approach is this mess:
```rust
let rv = mp4parse_new(&io, &mut parser as *mut *mut _ as *mut *const _);
```
In the end, it seemed worth switching to a slightly weirder signature, that had this much more ergonomic code for consumers:
```rust
        let mut rv = Mp4parseStatus::Invalid;
        let parser = mp4parse_new(&io, &mut rv);
```
Let me know if you disagree.

baumanj: No problem, I don't think. I guess I was being a bit of a stickler in my first attempt by making the `parser` out parameter `*mut *const` rather than `*mut *mut` to indicate that the caller passes in a pointer to be set, but that the dereferenced value should never be modified. Given that the other accessors *do* require mutation, for example:
```rust
pub unsafe extern fn mp4parse_get_track_info(parser: *mut Mp4parseParser, track_index: u32, info: *mut Mp4parseTrackInfo) -> Mp4parseStatus;
```
it's probably not worth it to make that distinction in `mp4parse_new`.

Let me try this other approach a bit and see if anything particularly ugly pops out. If not, agree with you that having the status be the return value is nicer since it's more consistent with the rest of the API.

PR #201, opened 2020-03-10
read_avif: Collapse primary_items_extents_data with Vec::concat instead of Iterator::flatten.

kinetiknz:
This improves the runtime of the public_avif_read_samples test from >20s to
<0.3s.  In optimized builds, there's no performance issue with
Iterator::flatten, so the runtime of both versions are the same.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -1338,7 +1338,7 @@ pub fn read_avif<T: Read>(f: &mut T, context: &mut AvifContext) -> Result<()> {
         }
     }
 
-    context.primary_item = primary_item_extents_data.into_iter().flatten().collect();
+    context.primary_item = primary_item_extents_data.concat();
 
     Ok(())
 }


baumanj:
I'm surprised at the >20s. Locally, when I run:
```
$ cargo clean
$ cargo build --tests
$ time cargo test public_avif_read_samples
```
With the existing code, I'm seeing output like
```
real	0m9.933s
real	0m8.545s
real	0m8.387s
```
and with the `concat` version:
```
real	0m1.678s
real	0m0.255s
real	0m0.242s
```

But either way, `concat` is much faster and simpler as well, so I think this is a good change!

In looking at this code, though, it occurs to me that either version will potentially allocate, and should be using some fallible version rather than stdlib `concat` (the same goes for `append` a few lines earlier), would you agree?

If so, I'll file a separate issue and address it shortly.

kinetiknz:
> In looking at this code, though, it occurs to me that either version will potentially allocate, and should be using some fallible version rather than stdlib `concat` (the same goes for `append` a few lines earlier), would you agree?
> 
> If so, I'll file a separate issue and address it shortly.

Sounds good, thanks.

baumanj:
Filed https://github.com/mozilla/mp4parse-rust/issues/204

PR #203, opened 2020-03-10
Refactor mp4parse_new to return Mp4parseStatus

baumanj:
https://github.com/mozilla/mp4parse-rust/pull/200#discussion_r390008980 was a good suggestion; I just didn't see it before merging https://github.com/mozilla/mp4parse-rust/pull/200, so here it is.

The Mp4parseParser is now set through an output parameter. Returning a status
is more consistent with the rest of the API and results in somewhat simpler code.

Equivalent changes for mp4parse_avif_new as well.

### DIFF
--- a/mp4parse_capi/src/lib.rs
+++ b/mp4parse_capi/src/lib.rs
@@ -22,9 +22,9 @@
 //!     read: Some(buf_read),
 //!     userdata: &mut file as *mut _ as *mut std::os::raw::c_void
 //! };
-//! let mut rv = mp4parse_capi::Mp4parseStatus::Invalid;
+//! let mut parser = std::ptr::null_mut();
 //! unsafe {
-//!     let parser = mp4parse_capi::mp4parse_new(&io, &mut rv);
+//!     let rv = mp4parse_capi::mp4parse_new(&io, &mut parser);
 //!     assert_eq!(rv, mp4parse_capi::Mp4parseStatus::Ok);
 //!     assert!(!parser.is_null());
 //!     mp4parse_capi::mp4parse_free(parser);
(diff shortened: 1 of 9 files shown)


mp4parse_capi/src/lib.rs
```diff
@@ -417,76 +417,66 @@ impl Read for Mp4parseIo {
...
+) -> Mp4parseStatus {
     // Validate arguments from C.
-    if io.is_null() || (*io).userdata.is_null() || (*io).read.is_none() || status_out.is_null() {
-        set_if_non_null(status_out, Mp4parseStatus::BadArg);
-        std::ptr::null_mut()
+    if io.is_null()
+        || (*io).userdata.is_null()
+        || (*io).read.is_none()
+        || parser_out.is_null()
+        || !(*parser_out).is_null()
```
kinetiknz: This check requires the caller to null-initialize the outparam before passing in the address, but it's fairly common in C languages to use the address of an uninitialized variable for outparams, e.g.

```c
Mp4ParseParser* parser;
int rv = mp4parse_new(&io, &parser);
```

I think it's better to permit this common pattern than to enforce pre-initialization, but as you've also added a test for if you'd prefer to keep this as-is, the documentation for mp4parse_{avif,}_new should mention it.

baumanj: My thinking here was that there's no way to tell whether a non-NULL passed-in value was merely uninitialized, or a valid pointer which should've been freed. If it's the latter case, calling this function will overwrite the value, possibly leaking. So, the value of possibly catching a leak early seemed to outweigh the cost of requiring callers to initialize to NULL.

That said, I did intend to document it, so thanks for catching that. I'll update shortly.

PR #205, opened 2020-03-10
Check that code conforms to rustfmt formatting

baumanj:


kinetiknz:
LGTM, thanks.

ChunMinChang:
BTW, `rustfmt` is not always available with _Rust Nightly_. The availability can be checked [here](https://rust-lang.github.io/rustup-components-history/).

When the `rustfmt` is unavailable in current _Rust Nightly_, it needs to install the latest available `rustfmt` to make `cargo fmt` work. Here is an [example](https://github.com/rust-lang/rust-clippy/blob/27acea0a1baac6cf3ac6debfdbce04f91e15d772/.travis.yml#L40-L46). 

Mmore detail can be found in [here](https://github.com/rust-lang/rustup-components-history/blob/master/README.md)

baumanj:
Maybe it would be easier to only do the check on stable? I need to fix an issue anyway since
```bash
cargo fmt -- --check || echo "some error message"
```
will always succeed due to the exit status of `echo`.

kinetiknz:
> BTW, `rustfmt` is not always available with _Rust Nightly_. The availability can be checked [here](https://rust-lang.github.io/rustup-components-history/).

That's a trap, thanks for the heads up!

> Maybe it would be easier to only do the check on stable?

Sounds good.

baumanj:
Fixes over here: https://github.com/mozilla/mp4parse-rust/pull/207

PR #206, opened 2020-03-10
Test rustfmt checking in CI [DO NOT MERGE]

baumanj:
This shouldn't be committed; it *should* fail CI.


baumanj:
Found a bug in https://github.com/mozilla/mp4parse-rust/pull/205

PR #207, opened 2020-03-10
Don't run rustfmt check in nightly; fix bug

baumanj:
There may be a cleaner way to do this with https://docs.travis-ci.com/user/conditions-v1, but I wasn't really sure, and I'm fairly confident this will do what we need.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,8 @@ env:
     - RELEASE=--release
 
 before_script:
-- rustup component add rustfmt
+# rustfmt may not always be available in nightly
+- if [[ "$TRAVIS_RUST_VERSION" != *nightly* ]]; then rustup component add rustfmt; fi
 
 script:
   - cargo test --all --verbose $RELEASE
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -14,7 +14,8 @@ env:
     - RELEASE=--release
 
 before_script:
-- rustup component add rustfmt
+# rustfmt may not always be available in nightly
+- if [[ "$TRAVIS_RUST_VERSION" != *nightly* ]]; then rustup component add rustfmt; fi
```
kinetiknz: I think $TRAVIS_RUST_VERSION can only ever be "stable", "beta", or "nightly" (from the .travis.yml docs), so the substring match isn't strictly needed AIUI.

baumanj: Yeah, I think you're right that it will use whatever literal is specified in the `.travis.yml`. Given that it would be technically possible to specify an older nightly there (e.g., `nightly-2020-03-01`, I'd say this approach would do more what we want.

If you think there's a _risk_ to this approach; I'm happy to change it. Otherwise, I think I'll leave it alone at this point.

baumanj:
Expected failure confirmed in https://github.com/mozilla/mp4parse-rust/pull/208. I think this is working correctly now.

PR #208, opened 2020-03-10
Test rustfmt checking in CI, take 2 [DO NOT MERGE]

baumanj:
https://github.com/mozilla/mp4parse-rust/pull/206 should've failed, but didn't.

That issue should be fixed by https://github.com/mozilla/mp4parse-rust/pull/207.


baumanj:
Failed (only on stable) as expected.

PR #209, opened 2020-03-11
Fix assert found by fuzzing + minor cleanups

kinetiknz:
- Rename fuzz_target_1
- Fix fuzz_target builds after PR #200 
- Fix assert found by fuzzing (see commit message for explanation)


kinetiknz:
> Is there any reason we can't add building the fuzz targets to CI? Without that, I imagine I'll forget and make build-breaking changes again.

I'll try that in a new PR.

PR #210, opened 2020-03-11
Build fuzz targets on CI.

kinetiknz:


kinetiknz:
x86-64 Linux and x86-64 macOS are currently only supported targets for cargo-fuzz.  However macOS is [failing](https://travis-ci.org/mozilla/mp4parse-rust/jobs/660896006) with https://github.com/rust-fuzz/cargo-fuzz/issues/217, so I'll exclude it for now.

Issue #204, opened 2020-03-10
Replace potentially allocating functions with fallible alternatives


baumanj:
There are already some instances of fallible alternatives like [`vec_push`](https://github.com/mozilla/mp4parse-rust/blob/0f883c3b5cf12580adb81a6c34676721081bf8ea/mp4parse/src/lib.rs#L128), [`vec_with_capacity`](vec_with_capacity), [`extend_from_slice`](https://github.com/mozilla/mp4parse-rust/blob/0f883c3b5cf12580adb81a6c34676721081bf8ea/mp4parse/src/lib.rs#L151), and [`read_to_end`](read_to_end). However, there are additional functions which may allocate such as [`Vec::append`](https://github.com/mozilla/mp4parse-rust/blob/0f883c3b5cf12580adb81a6c34676721081bf8ea/mp4parse/src/lib.rs#L1333) and [`slice::concat`](https://github.com/mozilla/mp4parse-rust/blob/0f883c3b5cf12580adb81a6c34676721081bf8ea/mp4parse/src/lib.rs#L1341) which need a similar treatment.

PR #211, opened 2020-03-12
Switch Vec to TryVec for fallible allocation

baumanj:
See https://github.com/mozilla/mp4parse-rust/issues/204

It's too easy to accidentally call functions in `Vec` which may allocate, so
rather than relying on the consumer to use `vec_push` instead of `Vec::push`
and the like, make `TryVec` a thin wrapper around `Vec`, which will return
`Result`s for all the operations which allocate and will use fallible allocation
when the `mp4parse_fallible` feature is enabled.

For now, as a proof of concept, only replace `Vec`s the newly-added AVIF
code. If this works well, expanding to the rest of the code shouldn't be
too much additional effort.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -174,7 +174,6 @@ mod fallible {
     use std::convert::TryInto as _;
     use std::io::Read;
     use std::io::Take;
-    use std::iter::FromIterator;
     use std::iter::IntoIterator;
     use vec_push;
     use vec_with_capacity;
@@ -229,6 +228,25 @@ mod fallible {
         inner: Vec<T>,
     }
 
+    /// TryVec is a thin wrapper around std::vec::Vec to provide support for
+    /// fallible allocation. Though the fallible allocator is only enabled
+    /// with the mp4parse_fallible feature, the API differences from stdlib
+    /// Vec ensure that all operations which allocate return a Result. For the
+    /// most part, this simply means adding a Result return value to Vec
+    /// functions which return nothing or a non-Result value. However, Vec
+    /// implements some traits whose API cannot communicate failure, but which
+    /// do require allocation, so it is important that TryVec does not
+    /// implement these traits.
+    ///
+    /// Specifically, do not implement any of the following traits:
+    /// - Clone
+    /// - Extend
+    /// - From
+    /// - FromIterator
+    ///
+    /// This list may not be exhaustive. Exercise caution when implementing
+    /// any new traits to ensure they won't potentially allocate in a way that
+    /// can't return a Result to indicate allocation failure.
     impl<T> TryVec<T> {
         pub fn new() -> Self {
             Self { inner: Vec::new() }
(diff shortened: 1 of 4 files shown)


mp4parse/src/lib.rs
```diff
@@ -160,32 +165,205 @@ pub fn extend_from_slice<T: Clone>(vec: &mut Vec<T>, other: &[T]) -> std::result
...
+    impl<T: Clone> TryVec<T> {
+        pub fn extend_from_slice(&mut self, other: &[T]) -> Result<(), ()> {
+            extend_from_slice(&mut self.inner, other)
+        }
+    }
+
+    impl<T: Clone> Clone for TryVec<T> {
+        fn clone(&self) -> Self {
+            Self {
+                inner: self.inner.clone(),
```
kinetiknz: This is tricky, since `Clone` has no way to signal failure, but the implied allocation in `self.inner.clone()` would ideally be fallible.

Can we get away with leaving this unimplemented and deliberately avoiding uses, or is that going to be too messy?

mp4parse/src/lib.rs
```diff
@@ -160,32 +165,205 @@ pub fn extend_from_slice<T: Clone>(vec: &mut Vec<T>, other: &[T]) -> std::result
...
+
+        fn deref(&self) -> &[T] {
+            self.inner.deref()
+        }
+    }
+
+    impl<T> FromIterator<T> for TryVec<T> {
+        fn from_iter<I: IntoIterator<Item = T>>(iter: I) -> Self {
+            Self {
+                inner: Vec::from_iter(iter),
```
kinetiknz: Same question as for `Clone`.

mp4parse/src/lib.rs
baumanj: Good catch! So far, it looks like we can get away with avoiding uses and I've updated the PR with a fix: see https://github.com/mozilla/mp4parse-rust/pull/211/commits/54aa1e594c9763e122bbb46be6fdabee24e493d5

I also noticed a similar bug in `extend_from_slice` which is fixed.

mp4parse/src/lib.rs
baumanj: See https://github.com/mozilla/mp4parse-rust/pull/211#discussion_r391393466

mp4parse/src/lib.rs
```diff
@@ -122,70 +120,310 @@ impl<'a, T: Read> Read for OffsetReader<'a, T> {
...
+                inner: self.inner.iter(),
+            }
+        }
+
+        pub fn pop(&mut self) -> Option<T> {
+            self.inner.pop()
+        }
+
+        pub fn push(&mut self, value: T) -> Result<()> {
+            self.reserve(if self.inner.capacity() == 0 { 4 } else { 1 })?;
```
baumanj: After pulling in this little bit of logic, the only remaining dependency on `FallibleVec` is `try_reserve`. I wonder if we should consider absorbing that into this repo.

kinetiknz: > I wonder if we should consider absorbing that into this repo.

:+1:

mp4parse_capi/src/lib.rs
```diff
@@ -62,8 +60,12 @@ use mp4parse::Track;
 use mp4parse::TrackScaledTime;
 use mp4parse::TrackTimeScale;
 use mp4parse::TrackType;
+use mp4parse::TryVec;
 use mp4parse::VideoCodecSpecific;
 
+#[allow(dead_code)]
+struct Vec; // To ensure we don't use any std::vec::Vec by accident
```
kinetiknz: Neat!

baumanj:
I'm going to do a try run on m-c with this before I merge.

mp4parse/src/lib.rs
baumanj: Will follow up on this with my next PR to handle the other potentially allocating types.

PR #213, opened 2020-03-17
Add types to support fallible allocation for String, HashMap and Box

baumanj:
Resolves https://github.com/mozilla/mp4parse-rust/issues/204

Also, add dummy zero-sized types to all the non-test code to help avoid
using the stdlib types by accident.


mp4parse/src/boxes.rs
```diff
@@ -70,23 +70,27 @@ impl From<BoxType> for FourCC {
...
+        FourCC { value: v }
     }
 }
 
 impl fmt::Debug for FourCC {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.value)
+        match std::str::from_utf8(&self.value) {
+            Ok(s) => write!(f, "{}", s),
+            Err(_) => self.value.fmt(f),
```
baumanj: This is a slight change, but I figure it's more useful to see the non-utf8 values in debugging rather than "null".

baumanj:
Per https://github.com/mozilla/mp4parse-rust/pull/211#discussion_r392046743, see https://github.com/mozilla/mp4parse-rust/pull/213/commits/9c659deced58f8ad3e00ce84d01ca220b25a532e.

mp4parse/src/boxes.rs
```diff
@@ -3,6 +3,16 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 use std::fmt;
 
+// To ensure we don't use stdlib allocating types by accident
```
SingingTree: Out of interest, is there a term for this strategy to prevent usage?

baumanj: Not to my knowledge. I suppose you could say it's a form of shadowing, but this is the first time I've ever _wanted_ to prevent the use of types that are in the stdlib (and most of them in the prelude).

mp4parse/src/fallible.rs
```diff
@@ -0,0 +1,513 @@
...
+//! any new traits to ensure they won't potentially allocate in a way that
+//! can't return a Result to indicate allocation failure.
+
+#[cfg(feature = "mp4parse_fallible")]
+extern "C" {
+    fn malloc(bytes: usize) -> *mut u8;
+    fn realloc(ptr: *mut u8, bytes: usize) -> *mut u8;
+}
+
+#[cfg(feature = "mp4parse_fallible")]
```
SingingTree: Can we drop this? It looks like we unconditionally need PartialEq? And will we always get it regardless of our config because of the prelude?

baumanj:
> LGTM. Could we benefit from testing (some of) the new types?

I'm open to that, but not sure what sort of stuff would be valuable since it's almost all just wrapping the stdlib types. There are a few in [`fallible.rs`](https://github.com/mozilla/mp4parse-rust/pull/213/files#diff-7a828b29d7077b37447850cf9db38681) (L341–386) that try to cover the most significant changes, but if you see other areas that you think could benefit from more test coverage, let me know and I'll see what I can come up with.

SingingTree:
For the other structs could we test to ensure they fail if given impossibly large allocations?

baumanj:
> For the other structs could we test to ensure they fail if given impossibly large allocations?

I poked at it a bit, and for `TryBox` I don't see a way to make it a useful test. Like stdlib `Box`, it allocates storage on the heap to take ownership of a stack value, and clearly we can't have a stack value impossibly large enough to trigger an OOM in heap allocation. Methods like `reserve` don't make sense for a box, which is of a known, fixed size. The best we could do is use something like [`new_uninit`](https://doc.rust-lang.org/stable/std/boxed/struct.Box.html#method.new_uninitl), but that's not available in stable rust anyway. I did try adding our own implementation of just the allocation portion:
```rust
    fn alloc<'a>() -> Result<&'a mut T> {
        let size = std::mem::size_of::<T>();
        unsafe {
            let ptr = malloc(size) as *mut T;
            ptr.as_mut().ok_or(super::Error::OutOfMemory)
        }
    }
```
but even that won't work because we'd still need to call it in our test like `TryBox::<[u8; std::usize::MAX]>::alloc()`, and that won't even build:
```
error: the type `[u8; 18446744073709551615]` is too big for the current architecture
```
So, if we wanted to expose something testable, it would have to be an allocating method that doesn't depend on the type and just takes a size. At that point, we're just writing a general allocator, not testing `TryBox`.

For `TryHashMap` we can do something like:
```rust
#[test]
#[cfg(feature = "mp4parse_fallible")]
fn tryhashmap_oom() {
    match TryHashMap::<char, char>::default().reserve(std::usize::MAX) {
        Ok(_) => panic!("it should be OOM"),
        _ => (),
    }
}
```

So, I'll add that, but I'm not seeing much else that would be useful. Do you?

SingingTree:
Makes sense. Cool by me.

baumanj:
I'll go ahead and merge this for now. I'm not going to update the the consumer code in firefox until https://bugzilla.mozilla.org/show_bug.cgi?id=1624698 is resolved, but that will likely be developed from the basis of this code, so I'd just as soon have this merged.

If you get a chance to look at this @kinetiknz and have any feedback, I'm happy to make any changes _post facto_.

Issue #218, opened 2020-04-10
Original format from ProtectionSchemeInfoBox (sinf) is not exposed via C API


SingingTree:
[This](https://github.com/mozilla/mp4parse-rust/blob/8239ceabc58902c10a338b6735f78176e37d7baf/mp4parse/src/lib.rs#L541) member contains the original format 4cc when dealing with encrypted streams. It would be useful to expose this to aid in handling encrypted media.

The specific case I have in mind is avc1 vs avc3. In avc3 an h264 stream is expected to contain certain information in band, meaning this information may be encrypted and unusable prior to decryption. By exposing the original format, consumers can better reason about what information should be available prior to decryption. In the Gecko use case this will help us understand if we can perform conversions to annex b or not.

PR #219, opened 2020-04-10
Expose encrypted original format via C API

SingingTree:
Fixes https://github.com/mozilla/mp4parse-rust/issues/218.

### DIFF
--- a/mp4parse/tests/public.rs
+++ b/mp4parse/tests/public.rs
@@ -309,7 +309,7 @@ fn public_audio_tenc() {
         assert_eq!(a.codec_type, mp4::CodecType::EncryptedAudio);
         match a.protection_info.iter().find(|sinf| sinf.tenc.is_some()) {
             Some(ref p) => {
-                assert_eq!(p.code_name, "mp4a");
+                assert_eq!(p.original_format.value, *b"mp4a");
                 if let Some(ref schm) = p.scheme_type {
                     assert_eq!(schm.scheme_type.value, *b"cenc");
                 } else {
@@ -368,7 +368,7 @@ fn public_video_cenc() {
         assert_eq!(v.codec_type, mp4::CodecType::EncryptedVideo);
         match v.protection_info.iter().find(|sinf| sinf.tenc.is_some()) {
             Some(ref p) => {
-                assert_eq!(p.code_name, "avc1");
+                assert_eq!(p.original_format.value, *b"avc1");
                 if let Some(ref schm) = p.scheme_type {
                     assert_eq!(schm.scheme_type.value, *b"cenc");
                 } else {
@@ -441,7 +441,7 @@ fn public_audio_cbcs() {
                 mp4::SampleEntry::Audio(ref a) => {
                     if let Some(p) = a.protection_info.iter().find(|sinf| sinf.tenc.is_some()) {
                         found_encrypted_sample_description = true;
-                        assert_eq!(p.code_name, "mp4a");
+                        assert_eq!(p.original_format.value, *b"mp4a");
                         if let Some(ref schm) = p.scheme_type {
                             assert_eq!(schm.scheme_type.value, *b"cbcs");
                         } else {
@@ -526,7 +526,7 @@ fn public_video_cbcs() {
                     assert_eq!(v.height, 300);
                     if let Some(p) = v.protection_info.iter().find(|sinf| sinf.tenc.is_some()) {
                         found_encrypted_sample_description = true;
-                        assert_eq!(p.code_name, "avc1");
+                        assert_eq!(p.original_format.value, *b"avc1");
                         if let Some(ref schm) = p.scheme_type {
                             assert_eq!(schm.scheme_type.value, *b"cbcs");
                         } else {
--- a/mp4parse/tests/public.rs
+++ b/mp4parse/tests/public.rs
@@ -309,9 +309,9 @@ fn public_audio_tenc() {
         assert_eq!(a.codec_type, mp4::CodecType::EncryptedAudio);
         match a.protection_info.iter().find(|sinf| sinf.tenc.is_some()) {
             Some(ref p) => {
-                assert_eq!(p.original_format.value, *b"mp4a");
+                assert_eq!(p.original_format, b"mp4a");
                 if let Some(ref schm) = p.scheme_type {
-                    assert_eq!(schm.scheme_type.value, *b"cenc");
+                    assert_eq!(schm.scheme_type, b"cenc");
                 } else {
                     panic!("Expected scheme type info");
                 }
@@ -368,9 +368,9 @@ fn public_video_cenc() {
         assert_eq!(v.codec_type, mp4::CodecType::EncryptedVideo);
         match v.protection_info.iter().find(|sinf| sinf.tenc.is_some()) {
             Some(ref p) => {
-                assert_eq!(p.original_format.value, *b"avc1");
+                assert_eq!(p.original_format, b"avc1");
                 if let Some(ref schm) = p.scheme_type {
-                    assert_eq!(schm.scheme_type.value, *b"cenc");
+                    assert_eq!(schm.scheme_type, b"cenc");
                 } else {
                     panic!("Expected scheme type info");
                 }
@@ -441,9 +441,9 @@ fn public_audio_cbcs() {
                 mp4::SampleEntry::Audio(ref a) => {
                     if let Some(p) = a.protection_info.iter().find(|sinf| sinf.tenc.is_some()) {
                         found_encrypted_sample_description = true;
-                        assert_eq!(p.original_format.value, *b"mp4a");
+                        assert_eq!(p.original_format, b"mp4a");
                         if let Some(ref schm) = p.scheme_type {
-                            assert_eq!(schm.scheme_type.value, *b"cbcs");
+                            assert_eq!(schm.scheme_type, b"cbcs");
                         } else {
                             panic!("Expected scheme type info");
                         }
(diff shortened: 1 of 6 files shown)


mp4parse/tests/public.rs
```diff
@@ -309,7 +309,7 @@ fn public_audio_tenc() {
         assert_eq!(a.codec_type, mp4::CodecType::EncryptedAudio);
         match a.protection_info.iter().find(|sinf| sinf.tenc.is_some()) {
             Some(ref p) => {
-                assert_eq!(p.code_name, "mp4a");
+                assert_eq!(p.original_format.value, *b"mp4a");
```
baumanj, line 312: It might be a boon to ergonomics to implement equality between `FourCC` and 4-byte arrays:
```rust
impl PartialEq<&[u8; 4]> for FourCC {
    fn eq(&self, other: &&[u8; 4]) -> bool {
        self.value.eq(*other)
    }
}
```
But since we only seem to use this for some tests, I'm fine with leaving this as is.

mp4parse_capi/src/lib.rs
```diff
@@ -220,6 +220,7 @@ pub struct Mp4parsePsshInfo {
 #[repr(C)]
 #[derive(Default, Debug)]
 pub struct Mp4parseSinfInfo {
+    pub original_format: Mp4parseByteData,
```
baumanj, line 223: Why not
```suggestion
    pub original_format: [u8; 4],
```
?

This simplifies the later assignments from
```rust
foo.original_format.set_data(&bar.value);
```
to
```rust
foo.original_format = bar.value;
```
and the tests from
```rust
        assert_eq!(protected_data.original_format.length, 4);
        for (i, expected_byte) in b"avc1".iter().enumerate() {
            assert_eq!(
                &(*protected_data.original_format.data.add(i)),
                expected_byte
            );
        }
```
to
```rust
        assert_eq!(protected_data.original_format, *b"avc1");
```
and (for the no encryption case)
```rust
        assert_eq!(protected_data.original_format, [0; 4]);
```

Probably the better way would be to make `original_format` `Option<[u8; 4]>`, but I think that poses issues for the FFI. I'm not sure if we have any shared code for mapping `Option<T>` to `mozilla::Maybe<T>`, but that would be nice.

If neither of those sound like improvements, the  `Mp4parseByteData` approach seems fine too.

mp4parse/tests/public.rs
SingingTree: Good call!

mp4parse_capi/src/lib.rs
SingingTree: I like this. As you say, the `Option` case doesn't straight forwardly seem to work with the FFI. I would prefer to represent this as optional as well, but am not sure of a nice way to do so with the bind gen. I'm open to ideas if you have any.

baumanj: I'm interested in the options for translating `Option` as well. I'll take a bit of time to investigate and let you know what I find.

baumanj: You're right that `Option` itself isn't FFI-safe, but I think it will work with something like:
```rust
#[repr(u8)]
#[derive(Debug, PartialEq)]
pub enum OptionalFormat {
    None,
    Some([u8; 4]),
}

impl Default for OptionalFormat {
    fn default() -> Self {
        Self::None
    }
}
```

I'm still trying it out to figure out the finer points and will follow up when I'm done.

Some relevant background, since I wasn't previously familiar with `repr(Int)`: https://github.com/rust-lang/rfcs/blob/master/text/2195-really-tagged-unions.md. Also, thanks to @emilio to pointing out this approach.

baumanj: Ok, I think it will work with the following changes:
```diff
diff --git a/mp4parse_capi/Cargo.toml b/mp4parse_capi/Cargo.toml
index 7326c98..c7d439c 100644
--- a/mp4parse_capi/Cargo.toml
+++ b/mp4parse_capi/Cargo.toml
@@ -35,7 +35,7 @@ num-traits = "0.2.0"
 env_logger = "0.7.1"
 
 [build-dependencies]
-cbindgen = "0.5.2"
+cbindgen = "0.14.1"
 
 [features]
 # Enable mp4parse_fallible to use fallible memory allocation rather than
diff --git a/mp4parse_capi/build.rs b/mp4parse_capi/build.rs
index b25eb1a..c31386f 100644
--- a/mp4parse_capi/build.rs
+++ b/mp4parse_capi/build.rs
@@ -34,6 +34,7 @@ extern "C" {
             "// THIS FILE IS AUTOGENERATED BY mp4parse_capi/build.rs - DO NOT EDIT".to_owned(),
         );
         c.language = cbindgen::Language::C;
+        c.cpp_compat = true;
         c.enumeration.rename_variants = Some(RenameRule::QualifiedScreamingSnakeCase);
         c
     };
diff --git a/mp4parse_capi/src/lib.rs b/mp4parse_capi/src/lib.rs
index e19c37a..f411aa5 100644
--- a/mp4parse_capi/src/lib.rs
+++ b/mp4parse_capi/src/lib.rs
@@ -217,10 +217,23 @@ pub struct Mp4parsePsshInfo {
     pub data: Mp4parseByteData,
 }
 
+#[repr(u8)]
+#[derive(Debug, PartialEq)]
+pub enum OptionalFormat {
+    None,
+    Some([u8; 4]),
+}
+
+impl Default for OptionalFormat {
+    fn default() -> Self {
+        Self::None
+    }
+}
+
 #[repr(C)]
 #[derive(Default, Debug)]
 pub struct Mp4parseSinfInfo {
-    pub original_format: Mp4parseByteData,
+    pub original_format: OptionalFormat,
     pub scheme_type: Mp4ParseEncryptionSchemeType,
     pub is_encrypted: u8,
     pub iv_size: u8,
@@ -828,10 +841,8 @@ fn get_track_audio_info(
             .iter()
             .find(|sinf| sinf.tenc.is_some())
         {
-            sample_info
-                .protected_data
-                .original_format
-                .set_data(&p.original_format.value);
+            sample_info.protected_data.original_format =
+                OptionalFormat::Some(p.original_format.value);
             sample_info.protected_data.scheme_type = match p.scheme_type {
                 Some(ref scheme_type_box) => {
                     match scheme_type_box.scheme_type.value.as_ref() {
@@ -992,10 +1003,8 @@ fn mp4parse_get_track_video_info_safe(
             .iter()
             .find(|sinf| sinf.tenc.is_some())
         {
-            sample_info
-                .protected_data
-                .original_format
-                .set_data(&p.original_format.value);
+            sample_info.protected_data.original_format =
+                OptionalFormat::Some(p.original_format.value);
             sample_info.protected_data.scheme_type = match p.scheme_type {
                 Some(ref scheme_type_box) => {
                     match scheme_type_box.scheme_type.value.as_ref() {
diff --git a/mp4parse_capi/tests/test_encryption.rs b/mp4parse_capi/tests/test_encryption.rs
index 277d8b7..66bbde2 100644
--- a/mp4parse_capi/tests/test_encryption.rs
+++ b/mp4parse_capi/tests/test_encryption.rs
@@ -51,13 +51,10 @@ fn parse_cenc() {
         assert_eq!((*video.sample_info).image_width, 320);
         assert_eq!((*video.sample_info).image_height, 240);
         let protected_data = &(*video.sample_info).protected_data;
-        assert_eq!(protected_data.original_format.length, 4);
-        for (i, expected_byte) in b"avc1".iter().enumerate() {
-            assert_eq!(
-                &(*protected_data.original_format.data.add(i)),
-                expected_byte
-            );
-        }
+        assert_eq!(
+            protected_data.original_format,
+            OptionalFormat::Some(*b"avc1")
+        );
         assert_eq!(
             protected_data.scheme_type,
             Mp4ParseEncryptionSchemeType::Cenc
@@ -86,13 +83,10 @@ fn parse_cenc() {
         assert_eq!((*audio.sample_info).bit_depth, 16);
         assert_eq!((*audio.sample_info).sample_rate, 44100);
         let protected_data = &(*audio.sample_info).protected_data;
-        assert_eq!(protected_data.original_format.length, 4);
-        for (i, expected_byte) in b"mp4a".iter().enumerate() {
-            assert_eq!(
-                &(*protected_data.original_format.data.add(i)),
-                expected_byte
-            );
-        }
+        assert_eq!(
+            protected_data.original_format,
+            OptionalFormat::Some(*b"mp4a")
+        );
         assert_eq!(protected_data.is_encrypted, 0x01);
         assert_eq!(protected_data.iv_size, 16);
         assert_eq!(protected_data.kid.length, 16);
@@ -142,13 +136,10 @@ fn parse_cbcs() {
         assert_eq!((*video.sample_info).image_width, 400);
         assert_eq!((*video.sample_info).image_height, 300);
         let protected_data = &(*video.sample_info).protected_data;
-        assert_eq!(protected_data.original_format.length, 4);
-        for (i, expected_byte) in b"avc1".iter().enumerate() {
-            assert_eq!(
-                &(*protected_data.original_format.data.add(i)),
-                expected_byte
-            );
-        }
+        assert_eq!(
+            protected_data.original_format,
+            OptionalFormat::Some(*b"avc1")
+        );
         assert_eq!(
             protected_data.scheme_type,
             Mp4ParseEncryptionSchemeType::Cbcs
@@ -207,11 +198,7 @@ fn parse_unencrypted() {
         assert_eq!(rv, Mp4parseStatus::Ok);
         assert_eq!(audio.sample_info_count, 1);
         let protected_data = &(*audio.sample_info).protected_data;
-        assert_eq!(protected_data.original_format.length, 0);
-        assert_eq!(
-            protected_data.scheme_type,
-            Mp4ParseEncryptionSchemeType::None
-        );
+        assert_eq!(protected_data.original_format, OptionalFormat::None);
         assert_eq!(protected_data.is_encrypted, 0x00);
         assert_eq!(protected_data.iv_size, 0);
         assert_eq!(protected_data.kid.length, 0);
```

It might also be worth adding a bit of test code in https://github.com/mozilla/mp4parse-rust/blob/master/test_ffi/src/test.cc to make sure the C++ side of things works like we'd expect.

baumanj:
Looks sane to me!

PR #223, opened 2020-05-13
Disable the ahash-compile-time-rng feature of hashbrown that makes bu…

glandium:
…ilds unreproducible.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1637140

### DIFF
--- a/mp4parse/Cargo.toml
+++ b/mp4parse/Cargo.toml
@@ -27,7 +27,8 @@ travis-ci = { repository = "https://github.com/mozilla/mp4parse-rust" }
 [dependencies]
 byteorder = "1.2.1"
 bitreader = { version = "0.3.2" }
-hashbrown = "0.7.1"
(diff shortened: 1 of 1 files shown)


kinetiknz:
LGTM, thanks!

Looks like this is also [fixed](https://github.com/rust-lang/hashbrown/commit/8858039cfa74443e49c364bd9932de5d418ffcac#diff-80398c5faae3c069e4e6aa2ed11b28c0) upstream in 0.7.2.

glandium:
Oh! So we actually don't need this, we can just update to hashbrown 0.7.2 in Firefox!

kinetiknz:
> Oh! So we actually don't need this, we can just update to hashbrown 0.7.2 in Firefox!

Sounds good, would you like me to revert this PR now?

glandium:
I don't have a preference.

PR #225, opened 2020-06-12
Improve debug format implementations

kornelski:
Using `{:?}` inside `Debug::fmt` implementation loses formatter flags, e.g. breaks `{:#?}` formatting.

I've also simplified other debug implementations where `write!` formatting machinery was unnecessary for printing a single string.

### DIFF
--- a/mp4parse/src/boxes.rs
+++ b/mp4parse/src/boxes.rs
@@ -44,7 +44,7 @@ macro_rules! box_database {
         impl fmt::Debug for BoxType {
             fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
                 let fourcc: FourCC = From::from(self.clone());
-                write!(f, "{}", fourcc)
+                fourcc.fmt(f)
             }
         }
     }
@@ -79,7 +79,7 @@ impl From<[u8; 4]> for FourCC {
 impl fmt::Debug for FourCC {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match std::str::from_utf8(&self.value) {
-            Ok(s) => write!(f, "{}", s),
+            Ok(s) => f.write_str(s),
             Err(_) => self.value.fmt(f),
         }
     }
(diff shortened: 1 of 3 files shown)


mp4parse/src/boxes.rs
```diff
@@ -79,18 +79,16 @@ impl From<[u8; 4]> for FourCC {
...
     }
 }
 
 impl fmt::Display for FourCC {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match std::str::from_utf8(&self.value) {
-            Ok(s) => write!(f, "{}", s),
-            Err(_) => write!(f, "null"),
-        }
+        let s = std::str::from_utf8(&self.value).ok().unwrap_or("null");
```
baumanj, line 90: Any reason not to make this
```suggestion
        let s = std::str::from_utf8(&self.value).unwrap_or("null");
```
?

kornelski: I've changed it to use same error handling as debug impl

baumanj: This ends up being a functional change though, yes? Quartets of bytes that aren't valid UTF-8 encodings which previously rendered as "null" when formatted via the `Display` trait will now print the underlying `[u8; 4]` values.

I'm not opposed to that change since it conveys more information, but the functionality predates my time with the project, so I'd want @kinetiknz to weigh in on whether that would be likely to cause any disruption. Considering we don't really expect to have any `FourCC` values which aren't valid UTF-8, the "null" stands out more as an erroneous condition.

In any case, if we do go this route, it makes the `Display` and `Debug` implementations identical, right? So we might as well do this:
```rust
impl fmt::Debug for FourCC {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        fmt::Display::fmt(self, f)
    }
}
```
or since there seems little point in implementing both identically, we could just derive `Debug` for `FourCC`.

kornelski: You're overthinking this and nitpicking tiny detail. This is a tiny function, I'm not sure if it's even used at all. Previous `null` was misleading, and the edge case is irrelevant anyway.

kornelski: I've reverted it to say `null`, and without redundant `.ok()`.

kinetiknz: We don't use Display/Debug for anything important (just dumping tools and other debugging), so it's fine to change to whatever makes the most sense now.

Displaying the raw `[u8; 4]` seems like a good change, it may be useful when debugging corrupt/unsupported boxes.

baumanj:
I'll merge this as-is for now. @kornelski, if you want to make a subsequent change to output the raw `[u8; 4]` instead of "null" when not valid UTF-8, that'd be fine too.

Thanks for the submission.

PR #228, opened 2020-07-30
If the iloc size is corrupt, return an Err rather than asserting

baumanj:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1655846


baumanj:
It looks like something is broken in the fuzzing CI, but that seems unrelated to this change as it started [sometime between 2 and 9 days ago](https://travis-ci.org/github/mozilla/mp4parse-rust/builds) without any changes to this repo.

I suggest we merge this change and investigate the CI failures separately.

kinetiknz:
> It looks like something is broken in the fuzzing CI, but that seems unrelated to this change as it started [sometime between 2 and 9 days ago](https://travis-ci.org/github/mozilla/mp4parse-rust/builds) without any changes to this repo.

It's due to libfuzzer-sys 0.3.3, which seems to need very new ASAN headers to build.  I [pinned](https://github.com/mozilla/mp4parse-rust/commit/63325444ae3388599f2f222775eebdde4c2f9f30) our version to 0.3.2 for now and will revisit this later.

PR #229, opened 2020-08-05
Always enable fallible allocation for fuzz targets

baumanj:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1657244


kinetiknz:
Thanks for fixing this!

PR #237, opened 2020-08-24
Struct for more avif meta

kornelski:
I've started work on support for alpha channel in AVIF. It requires collecting info from a few boxes and loading more than primary item.

Currently the parser doesn't even have ability to load more than info about one item, so this is a first step to remedy that.

I've removed early loading of data in `mdat` parser, because it duplicated a chunk of logic that I'll need to refactor, and I don't want to refactor it twice. The amount of copies stayed the same, because I've also removed unnecessary `concat()`.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -1070,46 +1070,6 @@ impl<'a, T: Read> BMFFBox<'a, T> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


baumanj:
Thanks for the submission! I've started reviewing this, but I have some other things I need to prioritize at the moment, so I may not be able to finish right away.

kornelski:
I've improved that code in #239, so I'll close this one.

Issue #221, opened 2020-05-04
TryBox is unsound


kornelski:
`TryBox` assumes that `Box::drop` will use `libc::free` to deallocate memory. This is not true if a crate uses jemallocator, or any custom allocator.

This could be fixed with:

1. Not using `Box` for `inner`, but keeping the raw pointer, and implementing `Drop` for `TryBox` that calls `libc::free`

2. or using std allocator API to use the same allocator as `Box`

I think option 1 is the simplest and makes least assumptions about Rust inner workings.

There's a more theoretical issue of `TryBox` not checking alignment. `malloc` gives alignment for largest platform-native type, but Rust types might ask for bigger alignment. An `assert` would suffice for this implementation.

`std::mem::forget(std::mem::replace` can be simplified to `ptr::write`.

baumanj:
Thanks for catching this. Would you agree that switching to [the `fallible_collections` version of `TryBox`](https://github.com/vcombey/fallible_collections/blob/f30aee9711d1d3edbfd222bef8d210cd0d821a7f/src/boxed.rs) would resolve this since it uses the stdlib allocator API to provide fallible allocation?

kornelski:
Yes, that looks good.

Issue #220, opened 2020-04-22
Use `std::alloc` functions instead of `(re)alloc` from libc


SimonSapin:
https://github.com/mozilla/mp4parse-rust/blob/8239ceabc58902c10a338b6735f78176e37d7baf/mp4parse/src/fallible.rs#L22-L25

The use of `Vec::from_raw_parts` and `Box::from_raw` with pointers from libc’s `malloc` or `realloc` is incorrect, and may cause Undefined Behavior when the Rust standard library’s global allocator is not libc (such as on Windows, or in programs that use `#[global_allocator]` to override it).

Functions from the `std::alloc` module should be used instead:

* https://doc.rust-lang.org/std/alloc/fn.alloc.html
* https://doc.rust-lang.org/std/alloc/fn.realloc.html

See also https://github.com/mozilla/mp4parse_fallible/issues/5 and https://phabricator.services.mozilla.com/D71742

Issue #146, opened 2018-03-17
Use nightly's try_reserve when available


snf:
Fallible allocation API just went live in nightly and it's breaking the package because of a name conflict with mp4parse_fallible's try_reserve.
Reference: https://bugzilla.mozilla.org/show_bug.cgi?id=1446538

kinetiknz:
I'll wait for try_reserve to hit stable before using it.  I'll leave this issue open to track that.

In the mean time, I've worked around the compilation failure with rustc nightly in https://github.com/mozilla/mp4parse-rust/commit/ded91a1661daaf55515e572aadd8e92e6d71cbaf.

kinetiknz:
https://github.com/kinetiknz/mp4parse-rust/tree/issue146

But I'm still waiting for this feature to become stable.

kinetiknz:
As a follow up to this, I've archived the mozilla/mp4parse_fallible repository since we no longer use it.

PR #242, opened 2020-09-01
Switch to using fallible_collections crate

baumanj:
Resolves https://github.com/mozilla/mp4parse-rust/issues/146
Resolves https://github.com/mozilla/mp4parse-rust/issues/220
Resolves https://github.com/mozilla/mp4parse-rust/issues/221

Additionally, this gets us one step closer to [sharing one rust fallible allocation implementation in Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1624698), and should unblock fuzzing/asan builds which disable the `moz_memory` feature.


.travis.yml
```diff
@@ -31,7 +31,6 @@ script:
   # (see comment where the feature is defined in mp4parse_capi/Cargo.toml),
   # but we can at least check for changes behind features that would break the
   # build.
-  - cargo check --all --verbose $RELEASE --tests --all-features
```
kinetiknz, line 34: We can remove the comment above as well, since it's out of date with this change

kinetiknz (approved):

.travis.yml
baumanj: Good point! Will do.

baumanj: Done

PR #247, opened 2020-11-04
Misc cleanup

baumanj:
This PR contains various cleanups that shouldn't change behavior in any meaningful way. The different kinds are separated by commit for ease of review.


SingingTree (approved):
👍

PR #248, opened 2020-11-05
Check ipma validity and preallocate associations vec

baumanj:
This fixes a timeout with [the attached testcase](https://github.com/mozilla/mp4parse-rust/files/5491742/clusterfuzz-testcase-minimized-avif-5139142796902400.zip) run against the `mp4parse_capi` crate with:
```
cargo +nightly fuzz run avif clusterfuzz-testcase-minimized-avif-5139142796902400 -- -timeout=60
```
Resolves https://oss-fuzz.com/testcase-detail/5139142796902400

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -1629,21 +1629,136 @@ pub struct AssociatedProperty {
     pub property: ItemProperty,
 }
 
-const MAX_IPMA_ASSOCIATION_COUNT: u64 = u8::MAX as u64;
+/// An upper bound which can be used to check overflow at compile time
+trait UpperBounded {
+    const MAX: u64;
+}
 
-/// Parse an ItemPropertyAssociation box
+/// Implement type $name as a newtype wrapper around an unsigned int which
+/// implements the UpperBounded trait.
+macro_rules! impl_bounded {
+    ( $name:ident, $inner:ty ) => {
+        #[derive(Clone, Copy)]
+        pub struct $name($inner);
+
+        impl $name {
+            pub const fn new(n: $inner) -> Self {
+                Self(n)
+            }
+
+            #[allow(dead_code)]
+            pub fn get(self) -> $inner {
+                self.0
+            }
+        }
+
+        impl UpperBounded for $name {
+            const MAX: u64 = <$inner>::MAX as u64;
+        }
+    };
+}
+
+/// Implement type $name as a type representing the product of two unsigned ints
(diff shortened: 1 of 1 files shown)


SingingTree (approved):
LGTM. It would be useful to me when reading the code to have a comment around the new block saying something along the lines of "verify size of box is sane given parameters" + "done verifying sanity of box" so I know I can skip this block if I'm assuming happy path.

Edit: something like "check validity" then "got valid size, per-allocate" would be good too. I'll stop bikeshedding the wording :) Ping me if there's anything further I can help look over.

baumanj:
> LGTM. It would be useful to me when reading the code to have a comment around the new block saying something along the lines of "verify size of box is sane given parameters" + "done verifying sanity of box" so I know I can skip this block if I'm assuming happy path.

That's a good idea. In fact, I think I'll break it out into a separate function to make that totally clear. Will update this likely tomorrow.

> Ping me if there's anything further I can help look over.

I originally made all the arithmetic use the `checked_XXX` functions instead of regular operators, but I found that harder to read, and considering most of it can be proven to make overflow/div-by-zero impossible, I'd appreciate you validating that as well to be sure I'm not missing anything. Worst case, we crash since [we have overflow checks on](https://github.com/mozilla/mp4parse-rust/blob/master/Cargo.toml#L8), but I'd still prefer to avoid that.

SingingTree:
> 
> I originally made all the arithmetic use the `checked_XXX` functions instead of regular operators, but I found that harder to read, and considering most of it can be proven to make overflow/div-by-zero impossible, I'd appreciate you validating that as well to be sure I'm not missing anything. Worst case, we crash since [we have overflow checks on](https://github.com/mozilla/mp4parse-rust/blob/master/Cargo.toml#L8), but I'd still prefer to avoid that.

I agree, I cannot see a way in the current code that we can overflow or divide by zero. Because this is based on a spec that is unlikely to blow us up by wildly widening a datatype (or changing anything, for that matter) I think this is good (if it were going to change a lot, maybe bite the bullet and use checked arithmetic to save on us hand checking each time). We could add a comment saying 'if changing this code please verify that is cannot overflow or divide by zero' as another barrier.

baumanj:
I was interested to see if I could get the compiler to validate that none of the arithmetic could overflow, and I think I was successful. This may be a bit overkill, but since this sort of validation keeps coming up in fuzzing, I think this may be of broader use in future work.

SingingTree (approved):

PR #253, opened 2020-11-18
Avoid unnecessary copies of AVIF primary_item and alpha_item

baumanj:
When they're already contiguously present in an `mdat` buffer (even the same one), `AvifContext::primary_item()` and `::alpha_item()` can return slices rather than having to copy the (potentially large) buffers.

This results in a measured performance increase of ~50% for parsing a 4.7 MB AVIF (the largest in our test set):

```
$ git checkout master
$ cargo bench -- --save-baseline base
avif_largest            time:   [2.0990 ms 2.1352 ms 2.1727 ms]
```
```
$ git checkout avif-nocopy
$ cargo bench -- --save-baseline avif-nocopy
$ cargo bench -- --load-baseline avif-nocopy --baseline base
avif_largest            time:   [1.0077 ms 1.0257 ms 1.0449 ms]
                        change: [-52.649% -51.371% -50.022%] (p = 0.00 < 0.05)
                        Performance has improved.
```
Also, clean up some related code add add tests.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -20,7 +20,6 @@ use num_traits::Num;
 use std::convert::{TryFrom, TryInto as _};
 use std::io::Cursor;
 use std::io::{Read, Take};
-use std::ops::{Range, RangeFrom};
 
 #[macro_use]
 mod macros;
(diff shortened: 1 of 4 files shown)


mp4parse/src/lib.rs
```diff
@@ -1302,45 +1445,71 @@ pub fn read_avif<T: Read>(f: &mut T) -> Result<AvifContext> {
...
                 }
             }
+
             if !found {
                 return Err(Error::InvalidData(
-                    "iloc contains an extent that is not in mdat",
+                    "iloc contains an extent that is not in any mdat",
                 ));
             }
         }
```
baumanj, line 1500: I considered this alternative implementation. I'm curious which approach others find most readable:
```rust
        if loc.extents.len() == 1 {
            let extent = loc.extents.pop().expect("exactly one extent");
            for mdat in &item_storage {
                if mdat.get(&extent).is_some() {
                    *item = Some(AvifItem::Location(extent));
                    break;
                }
            }

            if item.is_none() {
                return Err(Error::InvalidData(
                    "iloc contains an extent that is not in any mdat",
                ));
            }
        } else {
            let mut found = false;
            let mut item_data = TryVec::new();

            for extent in loc.extents {
                // try to find an mdat which contains the extent
                for mdat in item_storage.iter_mut() {
                    if let Some(extent_slice) = mdat.get(&extent) {
                        item_data.extend_from_slice(extent_slice)?;
                        found = true;
                        break;
                    }
                }

                if !found {
                    return Err(Error::InvalidData(
                        "iloc contains an extent that is not in any mdat",
                    ));
                }
            }

            *item = Some(AvifItem::Data(item_data));
        }
```

kinetiknz: The version in the commit seems nice, but no strong preference.

kinetiknz (approved):

PR #258, opened 2020-12-07
Add an API exposing alpha_item and premultiplied_alpha

ChunMinChang:
Create an API to expose the `alpha_item` and `premultiplied_alpha` with `Mp4parseAvifParser`. This helps to add the alpha support in [BMO1654462](https://bugzilla.mozilla.org/show_bug.cgi?id=1654462)

### DIFF
--- a/mp4parse_capi/src/lib.rs
+++ b/mp4parse_capi/src/lib.rs
@@ -436,6 +436,14 @@ pub struct Mp4parseParser {
     video_track_sample_descriptions: TryHashMap<u32, TryVec<Mp4parseTrackVideoSampleInfo>>,
 }
 
+#[repr(C)]
+#[derive(Default)]
+pub struct AvifImage {
+    pub primary_item: Mp4parseByteData,
+    pub alpha_item: Mp4parseByteData,
+    pub premultiplied_alpha: bool,
+}
+
 /// A unified interface for the parsers which have different contexts, but
 /// share the same pattern of construction. This allows unification of
 /// argument validation from C and minimizes the surface of unsafe code.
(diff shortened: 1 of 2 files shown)


baumanj:
This looks fine. Though I can't think of a situation where we'd ever call `mp4parse_avif_get_primary_item` or `mp4parse_avif_get_alpha_item` without calling the other. So, maybe it would be worth rolling them into a single function? It's not worth it if you think it would make the code a lot worse, but considering we could potentially have pages with dozens or hundreds of AVIFs in the not too distant future, it seems worth avoiding an extra FFI call for each one.

ChunMinChang:
Thanks for the feedback. I don't have a preference on whether `mp4parse_avif_get_primary_item` and `mp4parse_avif_get_alpha_item` should be merged or not. I'll merge them and then update the PR.

ChunMinChang:
The usage of the new API becomes: https://phabricator.services.mozilla.com/D98950#C3418652NL510

ChunMinChang:
The reason for the test failure in Rust Nightly is mentioned in https://github.com/mozilla/mp4parse-rust/pull/260#issuecomment-742100843. It seems it's unrelated to the change here actually.

mp4parse_capi/src/lib.rs
```diff
@@ -1177,26 +1185,33 @@ fn mp4parse_get_track_video_info_safe(
...
-/// primary_item raw pointers passed into it. Callers should ensure the parser
-/// pointer points to a valid `Mp4parseAvifParser`, and that the primary_item
-/// pointer points to a valid `Mp4parseByteData`. If there was not a previous
+/// avif_image raw pointers passed into it. Callers should ensure the parser
+/// pointer points to a valid `Mp4parseAvifParser`, and that the avif_image
+/// pointer points to a valid `AvifImage`. If there was not a previous
 /// successful call to `mp4parse_avif_read()`, no guarantees are made as to
-/// the state of `primary_item`.
+/// the state of `avif_image`. If `avif_image.alpha_item` are set with a
+/// positive`length` and non-null `data`, then the `avif_image` contains an
```
baumanj, line 1193:
```suggestion
/// the state of `avif_image`. If `avif_image.alpha_item` is set to a
/// positive `length` and non-null `data`, then the `avif_image` contains an
```
Just some typos. Overall, I really like this documentation update.

mp4parse_capi/src/lib.rs
```diff
@@ -1177,26 +1185,33 @@ fn mp4parse_get_track_video_info_safe(
...
+    let image_ref = &mut (*avif_image);
 
     let context = (*parser).context();
 
-    (*primary_item).set_data(context.primary_item());
+    image_ref.primary_item.set_data(context.primary_item());
+    if let Some(context_alpha_item) = context.alpha_item() {
+        image_ref.alpha_item.set_data(context_alpha_item);
+        image_ref.premultiplied_alpha = context.premultiplied_alpha;
+    }
```
baumanj, line 1214: For my own curiosity, is there any reason you added the `image_ref` binding as opposed to using `avif_image` directly like so?
```rust
    (*avif_image).primary_item.set_data(context.primary_item());
    if let Some(context_alpha_item) = context.alpha_item() {
        (*avif_image).alpha_item.set_data(context_alpha_item);
        (*avif_image).premultiplied_alpha = context.premultiplied_alpha;
    }
```
This works perfectly fine, so I'm happy to merge it. It just threw me for a bit reading it, so I was interested in the rationale.

baumanj (changes requested):
Happy to merge this after the typos in the doc comment are fixed. The functionality looks great. I like the introduction of the `AvifImage` type.

mp4parse_capi/src/lib.rs
ChunMinChang: I just worry the `*avif_image` would create a temporary variable cloned from `avif_image`. But I guess this won't happen in Rust. I wrote too much C++.

mp4parse_capi/src/lib.rs
ChunMinChang: Will fix the typoe. Thanks for the check!

mp4parse_capi/src/lib.rs
baumanj: Ah, that makes sense. I have trouble switching my mental model back and forth too, but my default is rust. Thanks for the explanation!

PR #259, opened 2020-12-07
Add an avif alpha test

ChunMinChang:
This test can also help to check different avif files by modifying the `IMAGE_AVIF_ALPHA` path


baumanj (approved):
Looks good!

Do you you have plans to add a non-premultiplied test case as well?

PR #260, opened 2020-12-09
Fix doctest failure

ChunMinChang:
This fixes the failure here: https://travis-ci.org/github/mozilla/mp4parse-rust/jobs/748610958#L725 in Rust Nightly.

### DIFF
--- a/mp4parse_capi/src/lib.rs
+++ b/mp4parse_capi/src/lib.rs
@@ -16,8 +16,8 @@
 //!        Err(_) => -1,
 //!    }
 //! }
-//!
-//! let mut file = std::fs::File::open("mp4parse/tests/minimal.mp4").unwrap();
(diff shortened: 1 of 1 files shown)


ChunMinChang:
Oops, it looks like this breaks Rust stable.

https://github.com/mozilla/mp4parse-rust/blob/fb4b4b97c0403e15673736ddfb310d6cd094c04f/mp4parse_capi/src/lib.rs#L20

The problem is that the `pwd` for the above call in Rust stable is `<path>/<to>/mp4parse-rust/mp4parse_capi`, while it's `<path>/<to>/mp4parse-rust/` in Rust Nightly.

kinetiknz:
The current working directory changing in nightly looks related to the recently fixed cargo issue 8097.

`CARGO_MANIFEST_DIR` seems like a fine solution for now.

Issue #261, opened 2020-12-12
c196ca8 cause MP4Metadata.test_case_mp4 fails


ChunMinChang:
`MP4Metadata.test_case_mp4` fails when updating `mp4parse-rust` crate. See failure here: https://treeherder.mozilla.org/logviewer?job_id=324286421&repo=autoland&lineNumber=29156 in [BMO 1681811](https://bugzilla.mozilla.org/show_bug.cgi?id=1681811).The first bad commit is commit c196ca8. The changes are made in #252. Now it blocks the avif-transparency support ([BMO 1654462](https://bugzilla.mozilla.org/show_bug.cgi?id=1654462)).

The error occurs in [these lines](https://searchfox.org/mozilla-central/rev/0bcf81557b89e7757c44e25bb4bc7f4cb8619dc9/dom/media/gtest/mp4_demuxer/TestParser.cpp#280,283,336). The test file is [`test_case_1185230.mp4`](https://searchfox.org/mozilla-central/rev/0bcf81557b89e7757c44e25bb4bc7f4cb8619dc9/dom/media/gtest/mp4_demuxer/TestParser.cpp#187). The error indicates [`MP4Metadata::GetNumberTracks`](https://searchfox.org/mozilla-central/rev/0bcf81557b89e7757c44e25bb4bc7f4cb8619dc9/dom/media/gtest/mp4_demuxer/TestParser.cpp#280,283,336) and [`MP4Metadata::GetTrackInfo`](https://searchfox.org/mozilla-central/rev/0bcf81557b89e7757c44e25bb4bc7f4cb8619dc9/dom/media/gtest/mp4_demuxer/TestParser.cpp#325,331,336) don't work as what it's expected. The underlying calls used in `MP4Metadata::GetNumberTracks` and `MP4Metadata::GetTrackInfo` are `mp4parse_get_track_count`, `mp4parse_get_track_info`, `mp4parse_get_track_audio_info`, and `mp4parse_get_track_video_info`. All the errors are around `track`.

I am not familiar with mp4 stuff enough to identify the problem quickly. But I guess there might be some mistakes around `read_mp4` and `read_moov`. It seems the track-info is loaded there. The `context` is changed from an input `&mut MediaContext` to an output `Result<MediaContext>`. Not sure if it causes the problem. Any idea?

baumanj:
See https://github.com/mozilla/mp4parse-rust/pull/263

PR #263, opened 2020-12-12
Restore read_moov behavior allowing multiple moov boxes

baumanj:
Resolves https://github.com/mozilla/mp4parse-rust/issues/261

https://treeherder.mozilla.org/jobs?repo=try&revision=96f98a79084b97402f73c8d3155d5889bdd75d35 is testing this applied to the mozilla-central tree and looks good. (Note that the [`ImageDecoders.AVIFStackCheck` error](https://bugzilla.mozilla.org/show_bug.cgi?id=1675073) on Windows is expected)

The behavior change was introduced by c196ca8e47a5df439b69c17df6ebd1ed59ed053d

Despite being invalid according to ISOBMFF (ISO 14496-12:2015) § 8.2.1.1, this restores the original behavior, while maintaining the interface that eliminates the necessity for callers to create a `MediaContext` object to pass in.

A brief note about the approach here:
[The original code for `read_moov`](https://github.com/mozilla/mp4parse-rust/blob/379151c6cb9b560844bee140e9e960b49149cb3e/mp4parse/src/lib.rs#L2112-L2146) took a `&mut MediaContext` parameter and returned a `Result<()>`. This was fairly natural when [the caller](https://github.com/mozilla/mp4parse-rust/blob/379151c6cb9b560844bee140e9e960b49149cb3e/mp4parse/src/lib.rs#L2046) (`read_mp4`) also took a `&mut MediaContext` parameter. The point of https://github.com/mozilla/mp4parse-rust/pull/252 was to improve ergonomics by replacing the in-out context parameters and `Result<()>` returns with functions that returned the context (inside a `Result`). There are a number of benefits to this approach. It's less code since
```rust
    let mut context = MediaContext::new();
    read_mp4(input)
```
is reduced to
```rust
    read_mp4(input)
```
and it eliminates the possibility that an empty/invalid context even exists to be operated on. However, in the case of `read_moov`, despite the fact that ISOBMFF (ISO 14496-12:2015) § 8.2.1.1 states there should be "exactly one" moov box at the file level, [at least one C++ test case](https://searchfox.org/mozilla-central/rev/95cf843de977805a3951f9137f5ff1930599d94e/dom/media/gtest/mp4_demuxer/TestParser.cpp#187) depends on the parser accepting a file with multiple moov boxes and merging their content. It's not clear to me if this is a concession to the reality of pervasive invalid MP4 files encoded this way in the wild, or if we just happened to create an invalid test case, but for the sake of not blocking other work, this restores the previous behavior, but with a slightly different approach.

The previous signature was
```rust
fn read_moov<T: Read>(f: &mut BMFFBox<T>, context: &mut MediaContext) -> Result<()>
```
and the new one is
```rust
fn read_moov<T: Read>(f: &mut BMFFBox<T>, context: Option<MediaContext>) -> Result<MediaContext>
```
Though this has the slightly unconventional approach of taking an owned `Option<MediaContext>` and returning a `Result<MediaContext>`, it has the benefit of not requiring the caller to ever create a `MediaContext` object, and since in the case of a `Some(_)` argument for `context`, that same object is returned, performance should be similar to using a `&mut MediaContext` parameter.


mp4parse/src/lib.rs
```diff
@@ -738,12 +738,6 @@ pub struct MediaContext {
     pub userdata: Option<Result<UserdataBox>>,
 }
 
-impl MediaContext {
-    pub fn new() -> MediaContext {
-        Default::default()
-    }
-}
-
```
baumanj, line 746: This could've been removed in https://github.com/mozilla/mp4parse-rust/commit/c196ca8e47a5df439b69c17df6ebd1ed59ed053d, but better late than never

mp4parse/tests/public.rs
```diff
@@ -618,6 +618,24 @@ fn public_video_av1() {
     }
 }
 
+#[test]
+fn public_mp4_bug_1185230() {
```
baumanj, line 622: I wanted to do more here, but it was turning into a yak shave, so I've filed https://github.com/mozilla/mp4parse-rust/issues/264

kinetiknz (approved):
I'm fine with reverting the behaviour, but it may be worth investigating whether we depend on this functionality for media in the wild.

>  It's not clear to me if this is a concession to the reality of pervasive invalid MP4 files encoded this way in the wild, or if we just happened to create an invalid test case, but for the sake of not blocking other work, this restores the previous behavior, but with a slightly different approach.

The original file attached to BMO bug 1185230 looks like the result of a fuzzing tool pasting a seed file together twice, so it seems to be an invalid testcase rather than a concession to media in the wild.  Splitting the file in half (offset 13651) and comparing the two halves reveals only 6 differing locations in sequences of 2, 8, 4, 4, 1, and 2 bytes - 3 changes inside an `mdat` binary blob, a key bug triggering change to the `time scale` of an `mdhd`  (3523726793 vs. 44100), and 2 inside a `data` element.

`test_case_1185230.mp4` seems to be a remuxed derivative of that file with `mdats`s removed and some other rearrangement, but I'm not sure exactly how it was produced.

mp4parse/src/lib.rs
```diff
@@ -2278,7 +2272,7 @@ pub fn read_mp4<T: Read>(f: &mut T) -> Result<MediaContext> {
                 debug!("{:?}", ftyp);
             }
             BoxType::MovieBox => {
-                context = Some(read_moov(&mut b)?);
+                context = Some(read_moov(&mut b, context)?);
```
ChunMinChang, line 2275: I guess the passed `context` here solves the problem?

ChunMinChang (approved):
I am not actively working on these stuff. But I am happy to ship it to unblock the AVIF work.

SingingTree (approved):
Makes sense to unblock work. Echoing what has already been said, if we can test that we don't need this in the wild it would be nice to not support it, but that's a task for telemetry in another patch.

baumanj:
For the sake of unblocking the AVIF work, I'm merging this for now. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1682376 to consider changing the behavior in the future.

PR #269, opened 2021-01-21
Make userdata parse failure non fatal

### LINKED ISSUE #268: Failure to parse itunes metadata breaks file parsing.
See https://bugzilla.mozilla.org/show_bug.cgi?id=1687357 for the initial report against Gecko.

When parsing the file in the above bug, we fail to parse the copyright itunes metadata box (which does look malformed to me, but I'm not overly familiar with said metadata). This breaks parsing entirely, which is unexpected. When supported landed for the metadata the intention was [failing to parse this data would not break other parsing](https://github.com/mozilla/mp4parse-rust/pull/182#discussion_r303311772).

I think I should be able to cook up a patch and a test. @kinetiknz, do you have an opinion on treatment of this kind of bad data? Right now, without having got a patch ready, my intention is to skip the single problematic item, that way we get the rest of the metadata. However, if the metadata has inter-item relationships, perhaps it would be best to fail the whole list once we encounter an error (which looks like what the current code is trying to do).

SingingTree:
This intends to fix #268 and provide test coverage for such a case.


SingingTree:
Repushed to tidy some comments in initial test file.

kinetiknz (approved):
LGTM, thanks.

FYI: Travis is failing on a small formatting issue in the stable builds.

SingingTree:
Thanks for the heads up. Re-pushed with the missing comma to fix formatting.

PR #282, opened 2021-03-12
Add clippy to build action.

kinetiknz:
Inspired by https://github.com/mozilla/mp4parse-rust/pull/186, updated for GitHub Actions.

Running a separate step for nightly clippy is annoying, but non-nightly builds don't recognize `clippy::upper_case_acronyms` and as far as I understand it's not currently possible to disable lints via Cargo.toml or clippy.toml.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -4,6 +4,14 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+// `clippy::upper_case_acronyms` is a nightly-only lint as of 2021-03-15, so we
+// allow `clippy::unknown_clippy_lints` to ignore it on stable - but
+// `clippy::unknown_clippy_lints` has been renamed in nightly, so we need to
+// allow `renamed_and_removed_lints` to ignore a warning for that.
+#![allow(renamed_and_removed_lints)]
+#![allow(clippy::unknown_clippy_lints)]
+#![allow(clippy::upper_case_acronyms)]
+
 #[macro_use]
 extern crate log;
 
@@ -41,6 +49,8 @@ const TABLE_SIZE_LIMIT: u32 = 30 * 60 * 60 * 24 * 7;
 /// A trait to indicate a type can be infallibly converted to `u64`.
 /// This should only be implemented for infallible conversions, so only unsigned types are valid.
 trait ToU64 {
+    // Remove when https://github.com/rust-lang/rust-clippy/issues/6727 is resolved
+    #[allow(clippy::wrong_self_convention)]
     fn to_u64(self) -> u64;
 }
 
@@ -59,6 +69,8 @@ impl ToU64 for usize {
 /// A trait to indicate a type can be infallibly converted to `usize`.
 /// This should only be implemented for infallible conversions, so only unsigned types are valid.
 pub trait ToUsize {
+    // Remove when https://github.com/rust-lang/rust-clippy/issues/6727 is resolved
+    #[allow(clippy::wrong_self_convention)]
     fn to_usize(self) -> usize;
 }
 
@@ -803,7 +815,7 @@ impl AvifContext {
                     "AvifItem::Location requires the location exists in AvifContext::item_storage"
                 );
             }
-            AvifItem::Data(data) => return data.as_slice(),
+            AvifItem::Data(data) => data.as_slice(),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


mp4parse/src/lib.rs
```diff
@@ -40,15 +41,15 @@ const TABLE_SIZE_LIMIT: u32 = 30 * 60 * 60 * 24 * 7;
 
 /// A trait to indicate a type can be infallibly converted to `u64`.
 /// This should only be implemented for infallible conversions, so only unsigned types are valid.
-trait ToU64 {
-    fn to_u64(self) -> u64;
+trait IntoU64 {
+    fn into_u64(self) -> u64;
```
baumanj, line 45: Why the change here? If there was a clippy issue about it, I can't seem to reproduce that. If it's a question of rust conventions, I think "to" is more idiomatic than "into" based on [the rust API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) for an owned -> owned conversion on `Copy` types.

mp4parse/src/lib.rs
```diff
@@ -2003,12 +2004,10 @@ fn read_ipma<T: Read>(
...
-            if previous_association.item_id > item_id {
-                return Err(Error::InvalidData(
-                    "Each ItemPropertyAssociation box shall be ordered by increasing item_ID",
-                ));
-            } else if previous_association.item_id == item_id {
-                return Err(Error::InvalidData("There shall be at most one association box for each item_ID, in any ItemPropertyAssociation box"));
+            match previous_association.item_id.cmp(&item_id) {
+                Ordering::Greater => return Err(Error::InvalidData( "Each ItemPropertyAssociation box shall be ordered by increasing item_ID")),
+                Ordering::Equal => return Err(Error::InvalidData("There shall be at most one association box for each item_ID, in any ItemPropertyAssociation box")),
+                Ordering::Less => {}
```
baumanj, line 2010: Personally, I find this less readable than the if/else version. I'd prefer adding an local exception or globally allowing this with a `-A clippy::comparison-chain` option. That said, if you feel strongly, I'm not about to die on this hill or anything.

mp4parse_capi/src/lib.rs
```diff
@@ -1514,10 +1506,7 @@ fn create_sample_table(
     };
 
     // According to spec, no sync table means every sample is sync sample.
-    let has_sync_table = match track.stss {
-        Some(_) => true,
-        _ => false,
-    };
+    let has_sync_table = matches!(track.stss, Some(_));
```
baumanj, line 1509: TIL about [`matches!`](https://doc.rust-lang.org/stable/core/macro.matches.html)

baumanj (commented):
Generally looks good, but I'd like to understand the rationale around the `To` -> `Into` naming change before signing off.

mp4parse/src/lib.rs
kinetiknz: I was working around https://github.com/rust-lang/rust-clippy/issues/6727, but I'll revert the change and add the appropriate clippy allow lines instead.

mp4parse/src/lib.rs
kinetiknz: I agree with you, I'll add an allow line for this instead.

kinetiknz:
I removed the nightly-only clippy CI step in favour of handling lint allow lines in the code itself - this has the advantage of in the generating the same results locally for a simple `cargo clippy` without having to copy the project-specific settings out of the CI config.

mp4parse/src/lib.rs
```diff
@@ -41,6 +49,7 @@ const TABLE_SIZE_LIMIT: u32 = 30 * 60 * 60 * 24 * 7;
 /// A trait to indicate a type can be infallibly converted to `u64`.
 /// This should only be implemented for infallible conversions, so only unsigned types are valid.
 trait ToU64 {
+    #[allow(clippy::wrong_self_convention)]
```
baumanj, line 52:
```suggestion
    // Remove when https://github.com/rust-lang/rust-clippy/issues/6727 is resolved
    #[allow(clippy::wrong_self_convention)]
```

mp4parse/src/lib.rs
```diff
@@ -59,6 +68,7 @@ impl ToU64 for usize {
 /// A trait to indicate a type can be infallibly converted to `usize`.
 /// This should only be implemented for infallible conversions, so only unsigned types are valid.
 pub trait ToUsize {
+    #[allow(clippy::wrong_self_convention)]
```
baumanj, line 71:
```suggestion
    // Remove when https://github.com/rust-lang/rust-clippy/issues/6727 is resolved
    #[allow(clippy::wrong_self_convention)]
```

baumanj (approved):
Looks good! Can we just add these comments referencing the appropriate clippy issue?

PR #291, opened 2021-04-05
Add support for XML metadata box

AdrianEddy:
This PR adds support for XML and Binary XML metadata boxes. 
They can be found for example in files coming from Sony cameras. 

Sample file here: https://drive.google.com/file/d/1EpOzytKGmx7YyKdmZW1_8b4pMHCVNGrs/view?usp=sharing

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -2362,7 +2362,7 @@ fn read_moov<T: Read>(f: &mut BMFFBox<T>, context: Option<MediaContext>) -> Resu
         mut mvex,
         mut psshs,
         mut userdata,
-        mut metadata,
+        metadata,
     } = context.unwrap_or_default();
 
     let mut iter = f.box_iter();
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -752,10 +752,19 @@ pub struct MetadataBox {
     pub sort_album_artist: Option<TryString>,
     /// The name of the composer to sort by 'soco'
     pub sort_composer: Option<TryString>,
+    /// Metadata
+    #[cfg(feature = "meta-xml")]
+    pub xml: Option<XmlBox>,
+}
+
+/// See ISOBMFF (ISO 14496-12:2015) § 8.11.2.1
+#[cfg(feature = "meta-xml")]
+#[derive(Debug)]
+pub enum XmlBox {
     /// XML metadata
-    pub xml: Option<TryString>,
+    StringXmlBox(TryString),
     /// Binary XML metadata
-    pub bxml: Option<TryVec<u8>>,
+    BinaryXmlBox(TryVec<u8>),
 }
 
 /// Internal data structures.
@@ -4062,7 +4071,8 @@ fn read_udta<T: Read>(src: &mut BMFFBox<T>) -> Result<UserdataBox> {
     Ok(udta)
 }
 
-/// Parse a metadata box inside a udta box
+/// Parse the meta box
+/// See ISOBMFF (ISO 14496-12:2015) § 8.111.
 fn read_meta<T: Read>(src: &mut BMFFBox<T>) -> Result<MetadataBox> {
     let (_, _) = read_fullbox_extra(src)?;
     let mut iter = src.box_iter();
@@ -4070,7 +4080,9 @@ fn read_meta<T: Read>(src: &mut BMFFBox<T>) -> Result<MetadataBox> {
     while let Some(mut b) = iter.next_box()? {
         match b.head.name {
             BoxType::MetadataItemListEntry => read_ilst(&mut b, &mut meta)?,
+            #[cfg(feature = "meta-xml")]
             BoxType::MetadataXMLBox => read_xml_(&mut b, &mut meta)?,
+            #[cfg(feature = "meta-xml")]
             BoxType::MetadataBXMLBox => read_bxml(&mut b, &mut meta)?,
             _ => skip_box_content(&mut b)?,
         };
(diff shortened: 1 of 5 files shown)


mp4parse/src/lib.rs
```diff
@@ -752,6 +752,10 @@ pub struct MetadataBox {
     pub sort_album_artist: Option<TryString>,
     /// The name of the composer to sort by 'soco'
     pub sort_composer: Option<TryString>,
+    /// XML metadata
+    pub xml: Option<TryString>,
+    /// Binary XML metadata
+    pub bxml: Option<TryVec<u8>>,
```
baumanj, line 758: Based on ISOBMFF (ISO/IEC 14496-12:2015) § 8.11.2.1 (emphasis mine):
> Quantity: Zero or one
…
> When the primary data is in XML format and it is desired that the XML be stored directly in the meta‐ box, **one of these forms may be used**.

I think making it structurally clear that only one or the other can be used is helpful:

```suggestion
    // metadata
    pub xml: Option<XmlBox>,
}

/// See ISOBMFF (ISO 14496-12:2015) § 8.11.2.1
#[derive(Debug)]
pub enum XmlBox {
    /// XML metadata
    StringXmlBox(TryString),
    /// Binary XML metadata
    BinaryXmlBox(TryVec<u8>),
}

```

mp4parse/src/lib.rs
```diff
@@ -4058,13 +4070,31 @@ fn read_meta<T: Read>(src: &mut BMFFBox<T>) -> Result<MetadataBox> {
...
         check_parser_state!(b.content);
     }
     Ok(meta)
 }
 
+/// Parse a XML box inside a meta box
+fn read_xml_<T: Read>(src: &mut BMFFBox<T>, meta: &mut MetadataBox) -> Result<()> {
+    let (_, _) = read_fullbox_extra(src)?;
+    let size = src.content.limit();
+    meta.xml = Some(read_buf(&mut src.content, size)?);
```
baumanj, line 4086: This can be done a bit more concisely using the convenience function in `fallible_collections` (and assuming the structure change I suggested above):
```suggestion
    meta.xml = Some(XmlBox::StringXmlBox(src.read_into_try_vec()?));
```

Similarly for `read_bxml`.

mp4parse/src/lib.rs
```diff
@@ -4058,13 +4070,31 @@ fn read_meta<T: Read>(src: &mut BMFFBox<T>) -> Result<MetadataBox> {
...
+            BoxType::MetadataXMLBox => read_xml_(&mut b, &mut meta)?,
+            BoxType::MetadataBXMLBox => read_bxml(&mut b, &mut meta)?,
             _ => skip_box_content(&mut b)?,
         };
         check_parser_state!(b.content);
     }
     Ok(meta)
 }
 
+/// Parse a XML box inside a meta box
```
baumanj, line 4082:
```suggestion
/// Parse a XML box inside a meta box
/// See ISOBMFF (ISO 14496-12:2015) § 8.11.2
```
Similarly for `read_bxml`.

mp4parse/src/lib.rs
```diff
@@ -4058,13 +4070,31 @@ fn read_meta<T: Read>(src: &mut BMFFBox<T>) -> Result<MetadataBox> {
...
+fn read_xml_<T: Read>(src: &mut BMFFBox<T>, meta: &mut MetadataBox) -> Result<()> {
+    let (_, _) = read_fullbox_extra(src)?;
```
baumanj, line 4084: Since the spec requires both the flags and version fields to be 0, I'd like this to validate that:
```suggestion
    if read_fullbox_version_no_flags(src)? != 0 {
        return Err(Error::Unsupported("unsupported XmlBox version"));
    }
```
Similarly for `read_bxml`.

baumanj (changes requested):
Thanks for the submission!

Overall, I think this looks great. Since it's feature work that's not required for mozilla-central (Firefox and other gecko-based projects), I'll ask you to put all this code behind a feature flag. See https://github.com/mozilla/mp4parse-rust/pull/280 for an example. I'd suggest `"meta-xml"` as the feature name, but if there's something else you like better, feel free.

I've suggested some small changes inline to be more consistent with the existing code. Let me know if you have any questions or concerns there.

Secondly, if you'd like to add some unit tests, that's fine, but since we won't run them in our CI, it's not required.

Finally, since this adds a call in a non-udta context for `read_meta`, let's update the doc comment (and add a spec reference while we're at it):
```
/// Parse the meta box
/// See ISOBMFF (ISO 14496-12:2015) § 8.111.
```

AdrianEddy:
@baumanj Thanks for the review, I updated the PR with requested changes and added the `meta-xml` feature. 

What do you think about the `metadata` field  in `MediaContext`? Right now we have it both in `MediaContext` and in `UserdataBox` so it follows the structure in mp4 files. But I think it can create confusion if someone expects the one in `MediaContext` to be present, while his file has it in `UserdataBox` or vice versa.

Should we set both metadata fields in each case or leave as is?

mp4parse/src/lib.rs
```diff
@@ -763,6 +776,7 @@ pub struct MediaContext {
     pub mvex: Option<MovieExtendsBox>,
     pub psshs: TryVec<ProtectionSystemSpecificHeaderBox>,
     pub userdata: Option<Result<UserdataBox>>,
+    pub metadata: Option<Result<MetadataBox>>,
```
baumanj, line 779:
```suggestion
    #[cfg(feature = "meta-xml")]
    pub metadata: Option<Result<MetadataBox>>,
```
This field is only used with the code added in this PR, so I think it belongs behind the feature flag as well.

mp4parse/src/lib.rs
```diff
@@ -2307,6 +2321,11 @@ pub fn read_mp4<T: Read>(f: &mut T) -> Result<MediaContext> {
             BoxType::MovieBox => {
                 context = Some(read_moov(&mut b, context)?);
             }
+            BoxType::MetadataBox => {
+                if let Some(ctx) = &mut context {
+                    ctx.metadata = Some(read_meta(&mut b));
+                }
+            }
```
baumanj, line 2328:
```suggestion
            #[cfg(feature = "meta-xml")]
                BoxType::MetadataBox => {
                if let Some(ctx) = &mut context {
                    ctx.metadata = Some(read_meta(&mut b));
                }
            }
```
Required based on putting `MediaContext.metadata` behind the feature flag

mp4parse/src/lib.rs
```diff
@@ -2352,6 +2371,7 @@ fn read_moov<T: Read>(f: &mut BMFFBox<T>, context: Option<MediaContext>) -> Resu
         mut mvex,
         mut psshs,
         mut userdata,
+        metadata,
```
baumanj, line 2374:
```suggestion
        #[cfg(feature = "meta-xml")]
        metadata,
```
Required based on putting `MediaContext.metadata` behind the feature flag

mp4parse/src/lib.rs
```diff
@@ -2394,6 +2414,7 @@ fn read_moov<T: Read>(f: &mut BMFFBox<T>, context: Option<MediaContext>) -> Resu
         mvex,
         psshs,
         userdata,
+        metadata,
```
baumanj, line 2417:
```suggestion
        #[cfg(feature = "meta-xml")]
        metadata,
```
Required based on putting `MediaContext.metadata` behind the feature flag

baumanj (changes requested):
Changes look great. I think we just need a few additional places put behind the feature flag and we're good to accept.

> What do you think about the `metadata` field in `MediaContext`? Right now we have it both in `MediaContext` and in `UserdataBox` so it follows the structure in mp4 files. But I think it can create confusion if someone expects the one in `MediaContext` to be present, while his file has it in `UserdataBox` or vice versa.
> 
> Should we set both metadata fields in each case or leave as is?

I think we leave it as it is. Though both cases use `meta` boxes, their intent is distinct.

Per ISOBMFF (ISO 14496-12:2015) § 8.11.2.1

> When the primary data is in XML format and it is desired that the XML be stored directly in the metabox, one of these forms may be used. The Binary XML Box may only be used when there is a single well‐defined binarization of the XML for that defined format as identified by the handler.

whereas § 8.10.1.1 defines the intent for `udta` boxes (emphasis mine):

> The User Data Box is a container box for *informative user‐data*. This user data is formatted as a set of boxes with more specific box types, which declare more precisely their content.

ISOBMFF is definitely complex and confusing, but I think there's a logic here and conflating these two use cases because both use `meta` boxes as containers would make it ultimately more confusing. I remain open to being convinced otherwise based on how specific handlers make use of this data, but for now, let's leave this as is.

baumanj (approved):

PR #293, opened 2021-04-06
Bump libfuzzer-sys dependency to 0.4.0

baumanj:
See https://github.com/google/oss-fuzz/issues/5572


kinetiknz (approved):
Thanks for fixing this.

SingingTree (approved):

Issue #133, opened 2017-11-30
How do I access track data?


fdoyon:
Hello, I am using this crate to read an mp4 container that contains audio tracks I need to decode, I am not sure how to get access to the audio packets of each individual track.

So far I can access the `SampleEntry::Audio` of each track but in order to decode the packets, I need to be able to calculate the offset each packet in the container file, which eludes me at the moment.

Thank you.

rillian:
Unfortunately this is not currently possible with just the Rust api in the `mp4parse crate`. You can do with the `Mp4Parse::sample_table_mut()` method from the `mp4parse_capi` crate.

It would be good to port that logic over, but our use in Firefox involves constructing `#[repr(C)]` structs we can pass to C++ code, so that's where we implemented it first.

alfredoyang:
Maybe we should move ```Mp4Parse::sample_table_mut``` to ```mp4parse crate``` and try to thin ```mp4parse_capi``` as much as possible.

palad1:
Thanks for the pointer, I was not sure what the use-case for ther `mp4parse_capi` crate and assumed it was only a thin C wrapper around `mp4parse` but it seems there's extra functionality which I should use.

PR #295, opened 2021-04-10
Make `create_sample_table` public

AdrianEddy:
This PR makes the `create_sample_table` function in `mp4parse_capi` public. It is needed to get the raw track data and looks like currently it's not possible to do it from Rust.

I tried to refactor both `mp4parse` and `mp4parse_capi` to move these functions to `mp4parse` crate, and leave `mp4parse_capi` as a thin C wrapper only (as mentioned in #133), but I've ran into issues with exposing structs and enums from `mp4parse` crate to cbindgen, so looks like it needs bigger refactoring than just moving functions around.

### DIFF
--- a/mp4parse_capi/src/lib.rs
+++ b/mp4parse_capi/src/lib.rs
@@ -1491,7 +1491,7 @@ impl<'a> SampleToChunkIterator<'a> {
 }
 
 #[allow(clippy::reversed_empty_ranges)]
-fn create_sample_table(
+pub fn create_sample_table(
     track: &Track,
     track_offset_time: CheckedInteger<i64>,
 ) -> Option<TryVec<Mp4parseIndice>> {


baumanj:
What's the specific use case here? `mp4parse_capi` exists to provide a FFI API and `create_sample_table` is not intended to be part of the public interface. If this is only necessary for access from rust code, then moving functionality to `mp4parse` and making an appropriate interface there is preferable.

If the problem with that approach is exposing `mp4parse` types from cbindgen, take a look at https://github.com/eqrion/cbindgen/issues/678, since an issue around that was just fixed.

AdrianEddy:
The use case is to read track raw data from file, and the result of this function provides offsets of samples. Without it, there is no way to do it from Rust, only from the C API. 

Yes, it will be more appropriate to refactor the entire capi, but I couldn't get cbindgen to play nice with it without bigger refactoring. Thanks for the info about eqrion/cbindgen#678, looks like this is what we need and I'll take another shot at refactoring the `mp4parse_capi`. (unless this is already being worked on?)

baumanj:
> The use case is to read track raw data from file, and the result of this function provides offsets of samples. Without it, there is no way to do it from Rust, only from the C API.

Ok, I think that sounds reasonable, but I'll need to see what the interface looks like before I can say we'll definitely accept it.

> Thanks for the info about [eqrion/cbindgen#678](https://github.com/eqrion/cbindgen/issues/678), looks like this is what we need and I'll take another shot at refactoring the `mp4parse_capi`. (unless this is already being worked on?)

No, it's not being worked on currently. It's fine to expose similar things from the `mp4parse` crate as we do from `mp4parse_capi`, and depending on the details, we can expand what the API allows access to. If your use case is for rust access, what types are you anticipating needing to move from `mp4parse_capi` to `mp4parse`?

AdrianEddy:
I'm mostly interested in moving the `create_sample_table` function and I managed to get cbindgen working with my changes, please take a look: https://github.com/mozilla/mp4parse-rust/compare/master...AdrianEddy:capi-refactor

This branch moves the functions and types which are not exposed and not using the `Mp4parseParser` C type, so they are also useful on the Rust side. I realize more types can be moved but I'm not sure if I should move them all at once or moving smaller parts like this is preferable.
There are no changes in the functions/types themselves, they are just moved around. One exception is `TrackType` type, which in C api has int values and lacks `Unknown` variant, and also the default is Video, while in mp4parse it's Unknown. What should we do with this one? And what value should `Unknown` have? 4?

If it's looking reasonable for you, I can create a new PR with these changes and we can close this one, and move the discussion to the new PR.

baumanj:
Thanks for the submission. Just to let you know, I'm not going to be able to start reviewing this until next week (Apr 19) at the earliest, but I'll definitely get to it as soon as I can.

AdrianEddy:
Ok, so I'll create a new PR with these changes to keep things organized

PR #296, opened 2021-04-12
Add support for MPEG-4 Part 2 video codec (ISO 14496-2).

dreamaniajp:
Hi, This PR tries to fix the MP4V  (ISO 14496-2) support.

Could you help to review it?
Many thanks!


baumanj:
Thanks for the submission. Just to let you know, I'm not going to be able to start reviewing this until next week (Apr 19) at the earliest, but I'll definitely get to it as soon as I can.

baumanj:
I've started looking at this, but I'm not going to be finished until later this week.

One thing that would help is to know what your desired use case for the code is. Are you looking to consume video/mp4v-es content via the `mp4parse` crate, or are you hoping to integrate it with some gecko-derived codebase? Something else?

dreamaniajp:
> I've started looking at this, but I'm not going to be finished until later this week.
> 
> One thing that would help is to know what your desired use case for the code is. Are you looking to consume video/mp4v-es content via the `mp4parse` crate, or are you hoping to integrate it with some gecko-derived codebase? Something else?

Hi,
I would like to integrate it with some gecko-derived codebase and would like to keep following/merging newer gecko-dev version in the future.
So it will be helpful if the video/mp4v-es support could be added into Gecko code base by some kind of feature flag control.

Many thanks.

baumanj:
It's going to take me a little while to get to this yet. Just wanted to let you know I haven't forgotten.

dreamaniajp:
> It's going to take me a little while to get to this yet. Just wanted to let you know I haven't forgotten.

Thanks @baumanj !

dreamaniajp:
Hi @baumanj , 

Sorry for bothering you.
If possible, could you help to have a look at this PR?

Thanks a lot!

baumanj:
Hey @dreamaniajp. Apologies for the delay, but this is a critical time for my main work. It would be easier to approve this PR if it were written in a way that doesn't affect the existing functionality when the `"mp4v"` feature isn't enabled. I understand that may be inconvenient, but since the way it's written now will potentially change the way the code behaves in Firefox, when  the feature is not enabled, I need to review it more thoroughly before we can accept it.

The current patch needs to have conflicts resolved (and probably be rebased). I'd suggest doing that and seeing if it's possible to write your change in a way that doesn't modify the existing functionality. We definitely want this crate to useful for purposes other than Firefox, but in terms of resource allocation, we have to prioritize that use case. If this is the approach you feel makes most sense, perhaps you can use a fork for the time being until I have the opportunity to give this the review it needs. I want to set clear expectations and let you know that it'll still be at least a little while until I can look at this closely enough to determine whether we can accept it in its current form.

dreamaniajp:
Hi @baumanj ,
Thanks for the reply, 
I rebased this PR and add "mp4v" feature to separate the mp4v related call flow from original.
This makes it doesn't affect the existing functionality when the "mp4v" feature isn't enabled. 

Thanks a lot.

mp4parse/src/lib.rs
```diff
@@ -3899,19 +3909,27 @@ fn read_dc_descriptor(data: &[u8], esds: &mut ES_Descriptor) -> Result<()> {
     let des = &mut Cursor::new(data);
     let object_profile = des.read_u8()?;
 
+    esds.audio_codec = match object_profile {
+        0x40 | 0x41 => CodecType::AAC,
+        0x69 | 0x6B => CodecType::MP3,
+        _ => CodecType::Unknown,
+    };
```
baumanj, line 3916: Is it necessary to move this? Doing so potentially changes the behavior even then the `"mp4v"` feature is not enabled since it mutates `esds` before the call to `find_descriptor` which takes it as a parameter

baumanj (changes requested):
Generally looks good, but I think there's one place where it is still potentially changing the existing behavior. Fix that one, and I think we should be good to go.

mp4parse/src/lib.rs
dreamaniajp: Hi @baumanj ,

No, we don't need to move this actually.
I moved this back to the original place to keep the original call flow and committed the change again.

Thanks.

mp4parse/src/lib.rs
```diff
@@ -3383,19 +3391,24 @@ fn read_dc_descriptor(data: &[u8], esds: &mut ES_Descriptor) -> Result<()> {
...
+
+    esds.video_codec = match object_profile {
+        0x20..=0x24 => CodecType::MP4V,
+        _ => CodecType::Unknown,
+    };
+
```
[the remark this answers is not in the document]
dreamaniajp: Hi @baumanj ,

I moved this back to the original place to keep the original call flow and committed the change again.

Thanks.

baumanj (approved):

fabricedesre:
Thanks all!

PR #297, opened 2021-04-12
Move functions which are not part of exposed C API from `mp4parse_capi` to `mp4parse` crate

AdrianEddy:
Move `create_sample_table`, `sample_to_chunk_iter`, `media_time_to_us`, `track_time_to_us` `Mp4parseIndice` and `Mp4parseTrackType` to `mp4parse` crate, as they are not part of the exposed C API and don't use `Mp4parseParser` C type.

This will allow users to access these functions from Rust.

There are no changes in the functions/types themselves, they are just moved around. One exception is `TrackType` type, which in C api has int values and lacks `Unknown` variant, and also the default is `Video`, while in mp4parse it's `Unknown`. What should we do with this one? And what value should Unknown have? 3? -1?


AdrianEddy:
Tests are failing, because the `Default` for `TrackType` changed from `Video` to `Unknown`, I think the correct default is `Unknown`. Should I update the test in `minimal_mp4_get_track_info_invalid_track_number` to expect `Unknown` track type? This seems logical, but it changes the behavior of the C api, so just wanted to make sure.

baumanj:
Will plan to take a look at this tomorrow. I just wanted to let you know it's still on my list

mp4parse/src/lib.rs
```diff
@@ -388,6 +392,529 @@ pub enum SampleEntry {
...
+}
+
+pub fn track_time_to_us<T>(time: TrackScaledTime<T>, scale: TrackTimeScale<T>) -> Option<T>
+where
+    T: PrimInt + Zero,
+{
+    assert_eq!(time.1, scale.1);
+    let microseconds_per_second = 1_000_000;
+    rational_scale::<T, u64>(time.0, scale.0, microseconds_per_second)
+}
```
baumanj, line 884: This is a move of https://github.com/mozilla/mp4parse-rust/blob/1ab30201ad06389d0d0ebe3036418d624e7219f4/mp4parse_capi/src/lib.rs#L707-L746 with the addition of making these three functions public.

Generally, the same points as with `create_sample_table` apply, so I think we could use the same approaches, but it's less clear to me why a consumer of this crate would want to expose *these* functions. Can you talk more about your use case?

Also, I think if we're going to make time conversion functions part of a public interface, it would be preferable to (rather than using `u64` or `T: PrimInt + Zero`) either:
- Make use of a newtype like `pub struct Microseconds<T>(T);` which makes unit-mixing mistakes easier to avoid
- Return a stdlib [Duration](https://doc.rust-lang.org/stable/std/time/struct.Duration.html) value

mp4parse/src/lib.rs
```diff
@@ -388,6 +392,529 @@ pub enum SampleEntry {
...
+
+#[test]
+fn track_time_overflow() {
+    let scale = TrackTimeScale(44100u64, 0);
+    let duration = TrackScaledTime(4_413_527_634_807_900u64, 0);
+    assert_eq!(
+        track_time_to_us(duration, scale),
+        Some(100_079_991_719_000_000)
+    );
+}
```
baumanj, line 916: This is a move of https://github.com/mozilla/mp4parse-rust/blob/1ab30201ad06389d0d0ebe3036418d624e7219f4/mp4parse_capi/src/lib.rs#L2088-L2118 without any modification. I appreciate moving the tests closer to the code they cover.

mp4parse/src/lib.rs
```diff
@@ -1123,11 +1650,12 @@ enum Extent {
...
 
+#[repr(C)]
 #[derive(Debug, PartialEq)]
 pub enum TrackType {
-    Audio,
-    Video,
-    Metadata,
+    Video = 0,
+    Audio = 1,
+    Metadata = 2,
```
baumanj, line 1658: To respond to [your comment](https://github.com/mozilla/mp4parse-rust/pull/297#issuecomment-818031114):
> Tests are failing, because the `Default` for `TrackType` changed from `Video` to `Unknown`, I think the correct default is `Unknown`. Should I update the test in `minimal_mp4_get_track_info_invalid_track_number` to expect `Unknown` track type? This seems logical, but it changes the behavior of the C api, so just wanted to make sure.

I don't think we should collapse `Mp4parseTrackType` into `TrackType`. The two are distinct, and I think that's intentional, because `TrackType` has the possibility of being the `Unknown` variant, whereas `Mp4parseTrackType` does not. Unless there's a valuable reason to expose that to the C API, I'd say it's better not to.

Do you have a use case which requires it?

baumanj (changes requested):
Overall, this looks good. A number of the comments I'm making here are just to reflect my validation of what exactly is changed in the big chunks of code which have moved between files. Those comments are more about my review due diligence than anything else, so they don't need any response unless I've misstated something.

The main issue that needs to be resolved before we can merge this is the question of what public interfaces these crates will expose and what expectations consumers can have about their stability. See inline comments [[1](https://github.com/mozilla/mp4parse-rust/pull/297#discussion_r619434552)], [[2](https://github.com/mozilla/mp4parse-rust/pull/297#discussion_r619470044)], [[3](https://github.com/mozilla/mp4parse-rust/pull/297#discussion_r619473376)] for more on that, but I think we should be able to figure something out.

mp4parse/src/lib.rs
AdrianEddy: Ok, you're right, `rational_scale` doesn't need to be public. The other two are public because they are used by `mp4parse_capi` crate. My usecase doesn't involve them directly (yet), but I do see them being useful for the consumer of this crate.

As for the return type, I think stdlib `Duration` is a better choice and I can go with that, but then should we also rename the functions? `track_time_to_duration`? or just leave as is?

mp4parse/src/lib.rs
```diff
@@ -388,6 +392,529 @@ pub enum SampleEntry {
...
+            if let [current_index, peek_index] = *indices {
+                let next_start_composition_time = sample_table[peek_index].start_composition;
+                let sample = &mut sample_table[current_index];
+                sample.end_composition = next_start_composition_time;
+            }
+        }
+    }
+
+    Some(sample_table)
+}
```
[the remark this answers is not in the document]
AdrianEddy: I added the `unstable-api` feature and also added the doc comments. Let me know if that is sufficient or I should make any changes.

BTW: Is there a cleaner way of adding `#[cfg(feature = "unstable-api")]`? Right now it's with every function, struct and import I moved and it doesn't look very clean in my opinion.

baumanj:
Thanks for the update! I should be able to take a look at this tomorrow or Wednesday, hopefully

baumanj:
Today is looking less likely, but hopefully by the end of the week.

baumanj:
I'm looking at this now and will hopefully have some additional feedback for you today, but in the meantime, I enabled the CI run and wanted to point out this one clippy failure:
```
error: this creates an owned instance just for comparison
    --> mp4parse/src/lib.rs:2593:25
     |
2593 |                 assert!(<$inner>::from(value) <= Self::MAX);
     |                         ^^^^^^^^^^^^^^^^^^^^^ help: try: `value`
...
2616 |     impl_bounded_product!(U32MulU8, U32, U8, u64);
     |     ---------------------------------------------- in this macro invocation
     |
     = note: `-D clippy::cmp-owned` implied by `-D warnings`
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cmp_owned
     = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

Isn't your issue, so don't worry about it.

mp4parse/src/lib.rs
baumanj: > As for the return type, I think stdlib `Duration` is a better choice and I can go with that, but then should we also rename the functions? `track_time_to_duration`? or just leave as is?

If it's not necessary to expose `media_time_to_us` and `track_time_to_us` for your use case, I'll leave it up to you to choose:
1. Keep them private
2. Change the return on the public (still behind `unstable-api` feature) functions to `Duration` (renaming them `track_time_to_duration` and `media_time_to_duration`), and leave `mp4parse_capi::{media_time_to_us, track_time_to_us}` as private functions, implemented in terms of the new `Duration`-returning ones.

mp4parse/src/lib.rs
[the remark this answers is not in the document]
baumanj: > BTW: Is there a cleaner way of adding `#[cfg(feature = "unstable-api")]`? Right now it's with every function, struct and import I moved and it doesn't look very clean in my opinion.

I'd suggest moving all that code to a new `unstable.rs` with
```
#![cfg(feature = "unstable-api")]
```
at the top. Then add
```
#[cfg(feature = "unstable-api")]
mod unstable;
```
to lib.rs. I think that should cut way down on how many places need `#[cfg(feature = "unstable-api")]`.

baumanj:
I think I responded to all your questions. Let me know if there's anything else you need, or else I'll take a look when you have another update.

mp4parse/src/lib.rs
AdrianEddy: Ok I went with newtype `Microseconds<T>` because stdlib `Duration` uses `u128` to represent micros and I didn't want to add that overhead

mp4parse/src/lib.rs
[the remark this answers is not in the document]
AdrianEddy: Moved to `unstable` mod, much cleaner now 👍

baumanj:
> I made all requested changes, please check if everything is good now

I'm in the process of giving this a last look over, but I have some things that come first, so it may be a few days yet. Just wanted to let you know I haven't forgotten

baumanj (approved):
This is plenty good enough to accept. There are a few tiny tweaks that I'd like your input on before merging, but if I don't hear back I'll likely merge this as is and possibly make those changes after the fact.

In the meantime, I'll just need to confirm this doesn't introduce any issue upstream in gecko that are discoverable with CI and then I'll merge.

mp4parse/src/unstable.rs
```diff
@@ -0,0 +1,547 @@
...
+    let remainder = numerator % denominator;
+    num_traits::cast(scale2).and_then(|s| match integer.checked_mul(&s) {
+        Some(integer) => remainder
+            .checked_mul(&s)
+            .and_then(|remainder| (remainder / denominator).checked_add(&integer)),
+        None => None,
+    })
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
```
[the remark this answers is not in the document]
AdrianEddy: Debug and PartialEq are needed for tests, otherwise this fails to compile: 
```
assert_eq!(
    track_time_to_us(duration, scale),
    Some(Microseconds(100_079_991_719_000_000u64))
);
```

mp4parse_capi/src/lib.rs
```diff
@@ -792,27 +621,31 @@ pub unsafe extern "C" fn mp4parse_get_track_info(
     let track = &context.tracks[track_index];
 
     if let (Some(track_timescale), Some(context_timescale)) = (track.timescale, context.timescale) {
-        let media_time: CheckedInteger<_> = match track.media_time.map_or(Some(0), |media_time| {
-            track_time_to_us(media_time, track_timescale)
-        }) {
-            Some(time) => time.into(),
+        let media_time: CheckedInteger<u64> = match track
```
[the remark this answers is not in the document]
AdrianEddy: No reason, it was an leftover from my testing, I changed it back now 👍

AdrianEddy:
Thanks for the review, I updated the code with all requested changes so it should be good to merge 🚀

baumanj:
Thanks for the contribution, the excellent work and your patience!

Issue #308, opened 2021-06-14
Require `utf8string` types to be null-terminated


baumanj:
Per ISOBMFF (ISO/IEC 14496-12:2020) § 4.2.1, fields of type `utf8string` are
> UTF-8 string as defined in IETF RFC 3629, null-terminated.

Currently this requirement is not enforced in the parsing of:
- `HandlerBox` (`hdlr`)
- `ItemInfoEntry` (`infe`)
- `SchemeTypeBox` (`schm`)

and possibly others

baumanj:
See https://github.com/mozilla/mp4parse-rust/pull/309#discussion_r651454804 for a link to code we've previously used to parse these `utf8string` types.

Given the invalid files in the wild we don't want to break, the approach here will likely be to be stricter* in the context of AVIF, since it's a newer format, but more permissive in other contexts.

Per the comment in old, deleted code:
```
    // XXX(kinetik): definition of "null-terminated" string is fuzzy, we have:
    // - zero or more byte strings, with a single null terminating the string.
    // - zero byte strings with no null terminator (i.e. zero space in the box for the string)
    // - length-prefixed strings with no null terminator (e.g. bear_rotate_0.mp4)
    // - multiple byte strings where more than one byte is a null.
```
Based on this text from ISOBMFF (ISO/IEC 14496-12:2020) § 4.2.1:
> In these definitions, null-terminated means that the last character of a string is Unicode NUL, and hence an empty string is represented by a single Unicode NUL.

I think `zero byte strings with no null terminator` are invalid (but we will continue to accept what we've accepted before). It's less clear what to do with `multiple byte strings where more than one byte is a null` since the spec only requires the last character to be NUL, not that the first NUL is necessarily the last character.

I've filed https://github.com/MPEGGroup/isobmff/issues/27 to request clarification in the spec.

\* That is, implement the standard

PR #309, opened 2021-06-14
Make hdlr box parsing reject invalid input according to ParseStrictness

baumanj:
See https://github.com/mozilla/mp4parse-rust/issues/308

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -204,6 +204,12 @@ impl From<std::string::FromUtf8Error> for Error {
     }
 }
 
+impl From<std::str::Utf8Error> for Error {
+    fn from(_: std::str::Utf8Error) -> Error {
+        Error::InvalidData("invalid utf8")
+    }
+}
+
 impl From<std::num::TryFromIntError> for Error {
     fn from(_: std::num::TryFromIntError) -> Error {
         Error::Unsupported("integer conversion failed")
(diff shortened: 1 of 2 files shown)


mp4parse/src/lib.rs
```diff
@@ -3081,7 +3087,7 @@ fn read_mdia<T: Read>(f: &mut BMFFBox<T>, track: &mut Track) -> Result<()> {
                 debug!("{:?}", mdhd);
             }
             BoxType::HandlerBox => {
-                let hdlr = read_hdlr(&mut b)?;
+                let hdlr = read_hdlr(&mut b, ParseStrictness::Permissive)?;
```
baumanj, line 3090: When parsing non-AVIF I left the behavior unchanged

kinetiknz: Thanks, looks good.  We used to read the component name field, but saw [various invalid forms](https://github.com/mozilla/mp4parse-rust/commit/012fd663126e8812aa2d0355b0823cf1e082df2c?branch=012fd663126e8812aa2d0355b0823cf1e082df2c&diff=unified#diff-f34d58d35412a7aa40d5fddc3be5b95878557f09b0618633c2654e1bf21ebfceL2001) in the wild and eventually dropped the code since we never used the field.

kinetiknz (approved):
LGTM

Issue #310, opened 2021-06-18
What's the purpose of `TABLE_SIZE_LIMIT`?


AdrianEddy:
I've ran into `OutOfMemory` errors when parsing a file. Besides this error being confusing, I tracked it down to checking the `TABLE_SIZE_LIMIT` for `stts` sample count. That file is a regular file from Sony RX0II camera, with a length of `6:23`. Audio track sample count is `18 402 480` with sample delta `1`, and this exceeded the `TABLE_SIZE_LIMIT` variable causing the `OutOfMemory` error.

I could just make a PR increasing this limit, but I'm pretty sure someone will run into this issue again sooner or later.
So my question is: What is the purpose of this limit? Why don't we just rely on TryVec and failing to allocate if we are indeed out of memory?

baumanj:
Based on [this PR](https://github.com/mozilla/mp4parse-rust/pull/97) and [this bugzilla issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1367508) it references, I believe it was introduced before we had support for fallible allocation in our collection types. That appears that [it was added a few months later](https://github.com/mozilla/mp4parse-rust/commit/c9a4a1c1912b23caf72e6a6f071aef489a48f24b).

At this point, we should be using [`fallible_collections`](https://github.com/vcombey/fallible_collections) for everything, so my guess is `TABLE_SIZE_LIMIT` can be removed, but this was before my time, so I'll defer to @kinetiknz, who was around then.

PR #311, opened 2021-06-21
Remove `TABLE_SIZE_LIMIT` and `BUF_SIZE_LIMIT`

kinetiknz:
Addresses https://github.com/mozilla/mp4parse-rust/issues/310

### DIFF
--- a/mp4parse/src/unstable.rs
+++ b/mp4parse/src/unstable.rs
@@ -161,9 +161,8 @@ pub fn create_sample_table(
     // With large numbers of samples, the cost of many allocations dominates,
     // so it's worth iterating twice to allocate sample_table just once.
     let total_sample_count = sample_to_chunk_iter(&stsc.samples, &stco.offsets)
-        .by_ref()
         .map(|(_, sample_counts)| sample_counts.to_usize())
-        .sum();
+        .try_fold(0usize, usize::checked_add)?;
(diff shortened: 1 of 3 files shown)


mp4parse/src/unstable.rs
```diff
@@ -161,9 +161,9 @@ pub fn create_sample_table(
     // With large numbers of samples, the cost of many allocations dominates,
     // so it's worth iterating twice to allocate sample_table just once.
     let total_sample_count = sample_to_chunk_iter(&stsc.samples, &stco.offsets)
-        .by_ref()
-        .map(|(_, sample_counts)| sample_counts.to_usize())
-        .sum();
+        .try_fold(0usize, |acc, (_, sample_counts)| {
+            acc.checked_add(sample_counts.to_usize())
+        })?;
```
baumanj, line 166: Good catch on the missing checked add!

Because I can't resist a good FP chain, I think this can be made just slightly more readable with the following refactor:
```suggestion
        .map(|(_, sample_counts)| sample_counts.to_usize())
        .try_fold(0, usize::checked_add)?;
```

baumanj (approved):
I love a good code removing PR!

This looks great, I just had one optional suggestion which you can implement or not according to whether you agree that it improves readability

mp4parse/src/unstable.rs
kinetiknz: That's tidier, thanks!

PR #312, opened 2021-06-22
Gracefully fail when the number of bytes a ctts box reports overflows `u32`

baumanj:
Resolves https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=35486


mp4parse/src/lib.rs
```diff
@@ -3469,7 +3467,7 @@ fn read_ctts<T: Read>(src: &mut BMFFBox<T>) -> Result<CompositionOffsetBox> {
         })?;
     }
 
-    skip_box_remain(src)?;
+    check_parser_state!(src.content);
```
baumanj, line 3470: I couldn't see a reason why there would ever legally be bytes remaining here and it didn't show up in any tests, but if there's a reason not to make this change, it's not necessary to fix the fuzzing issue.

kinetiknz (approved):
LGTM, thanks!

PR #326, opened 2021-08-05
Recognize ISO 13818-7 MPEG-2 object types for AAC.

kinetiknz:
Object types sourced from https://mp4ra.org/#/object_types.

This addresses [BMO 1722497](https://bugzilla.mozilla.org/show_bug.cgi?id=1722497).  I included `0x66` as well, since that's listed in the above reference as another (uncommon) AAC object type.

Curiously, the `0x41` object type we've accepted since this code was first added isn't listed.  I'm not sure where that value came from.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -4068,7 +4068,7 @@ fn read_dc_descriptor(data: &[u8], esds: &mut ES_Descriptor) -> Result<()> {
     }
 
     esds.audio_codec = match object_profile {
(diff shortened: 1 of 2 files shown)


kinetiknz:
Thanks for looking into `0x41` further - I wasn't able to find any reference (including in the old stagefright code) suggesting it should be treated as anything but a reserved OTI.  I've added a second commit to this PR to remove it.  Merging after the next nightly cycle sounds like a great plan.

PR #327, opened 2021-08-07
Add "missing-pixi-permitted" feature

baumanj:
Currently the `PixelInformationProperty` (`pixi`) box defined in HEIF (ISO 23008-12:2017) § 6.5.6 is required to be present per MIAF (ISO/IEC 23000-22:2019) specification § 7.3.6.6:
> The pixel information property shall be associated with every image that is displayable (not hidden) 

However due to a bug ([now fixed](https://github.com/strukturag/libheif/commit/f8e2306c8f803fc158e6decb32ff1c78d1f7eef8)) in [libheif](https://github.com/strukturag/libheif), AVIF images lacking this property were created and may remain in the wild. After receiving https://bugs.chromium.org/p/chromium/issues/detail?id=1198455, Chrome relaxed their check for this requirement and it is unclear whether it will be reinstated.

Discussion is underway to modify MIAF to allow the pixi property to be omitted, but it's unclear whether or when that will happen. To allow for maximum flexibility to adjust mp4parse's behavior, this PR adds a `"missing-pixi-permitted"` feature so the behavior can be finely tuned from within gecko without requiring an update of mp4parse code.

https://github.com/mozilla/mp4parse-rust/commit/47bff4cab7fa656acb2fe25a2bfbb16226a9e2cd makes the change assuming https://github.com/MPEGGroup/MIAF/issues/9 will be accepted. https://github.com/mozilla/mp4parse-rust/commit/891df7c59549c31aedeb908be89c82bea028f326 adds the feature and test changes that are only required during this period of uncertainty and can be removed later.
 
See also
- https://github.com/MPEGGroup/MIAF/issues/9
- https://github.com/mozilla/mp4parse-rust/issues/287
- https://github.com/AOMediaCodec/libavif/issues/564

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-    name: Build Rust ${{ matrix.rust }} on ${{ matrix.os }}, release=${{ matrix.release }}
+    name: Build Rust ${{ matrix.rust }} on ${{ matrix.os }}, release=${{ matrix.release }} features=${{ matrix.features }}
     runs-on: ${{ matrix.os }}
     continue-on-error: ${{ matrix.experimental }}
     strategy:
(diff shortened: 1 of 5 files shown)


kinetiknz (approved):
LGTM

PR #329, opened 2021-08-09
Add `cargo doc` to CI

baumanj:
This was present in the travis configuration that was removed in https://github.com/mozilla/mp4parse-rust/commit/f3c5f88d93da28c6c6dcd6d352d5ba9baab3b354. It remains useful since it can catch errors in the rustdoc such as in https://travis-ci.org/github/mozilla/mp4parse-rust/jobs/774702058


kinetiknz (approved):
Thanks!

I removed the Travis config thinking we'd fully migrated off of it - I should've looked more closely, sorry.

PR #332, opened 2021-08-18
Run clippy on fuzzer code

baumanj:
I checked this worked by intentionally re-introducing a warning for this run: https://github.com/mozilla/mp4parse-rust/runs/3366215594

It's now been reverted and everything should be green.


kinetiknz (approved):
Good idea, thanks!

PR #333, opened 2021-08-19
Expose the pixel information (pixi) via the C API

baumanj:
This is required for https://bugzilla.mozilla.org/show_bug.cgi?id=1696045, so that we can record telemetry on whether the `pixi` box is present


kinetiknz (approved):
LGTM, thanks.

PR #338, opened 2021-09-08
Add support for multiple `colr` box values per image item

baumanj:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1729071


mp4parse/src/lib.rs
```diff
@@ -2229,6 +2253,27 @@ fn read_iprp<T: Read>(
                                 )?;
                             }
                         }
+                        // XXX this is contrary to the published specification; see doc comment
```
SingingTree, line 2256: This is far enough from the doc comment on `read_iprp` that I'd consider explicitly saying 'see doc comment on read_iprp' or similar.

SingingTree (approved):
Looks good. Suggestion on one comment. I'm not familiar with how these colour boxes are used, but my assumption is I don't need to know that here -- the issue being fixed is the handling of multiple different types of box (and tolerance for multiple of the same type), not the later application of those boxes.

baumanj:
> the issue being fixed is the handling of multiple different types of box (and tolerance for multiple of the same type), not the later application of those boxes.

That is correct. The text of the upcoming change is not finalized (see https://github.com/AOMediaCodec/av1-avif/issues/164), so for now I'm handling the multiple boxes in a very permissive way, and even slightly ambiguous way, which will be tightened up once the specification is clear

Issue #339, opened 2021-09-09
Release a new version to crates.io


AdrianEddy:
Please release a new version to crates.io, there was a lot of useful changes since `v0.11.5` was released in January

baumanj:
Will get to this as soon as we can, and will plan to have a more consistent release policy going forward so stay tuned.

PR #347, opened 2021-10-20
Update version in anticipation of release on crates.io

baumanj:
This can be merged here now, but the release on github and crates shouldn't be published until Firefox 95 ships (December 7) so that the patch version matches the earliest version of the Firefox release it's used in. See README.md for more.

See https://github.com/mozilla/mp4parse-rust/issues/339


baumanj:
https://github.com/mozilla/mp4parse-rust/pull/355 has superceded this versioning-wise, and after further consideration, I think it makes more sense to indicate the particular version of the library used in a given Firefox release by adding tags ([for example](https://github.com/mozilla/mp4parse-rust/releases/tag/firefox-93)).

I will add the README.md note in a separate PR.

PR #358, opened 2021-12-18
Reserve space in pssh to avoid timeout

baumanj:
See https://oss-fuzz.com/testcase-detail/4609562076905472


kinetiknz (approved):
LGTM, thanks!

PR #364, opened 2021-12-29
Only read stsd entries up to `description_count`

baumanj:
Resolves https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=42930

Also, add some additional debugging statements I expect to to be useful in similar fuzzing issues


kinetiknz (approved):
LGTM!

PR #366, opened 2022-01-16
Upgrade to 2018 edition

kornelski:
This create already won't work with 2015-only versions of Rust, so it may as well update to a newer edition.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -35,7 +35,7 @@ use std::io::{Read, Take};
 mod macros;
 
 mod boxes;
-use boxes::{BoxType, FourCC};
+use crate::boxes::{BoxType, FourCC};
 
 // Unit tests.
 #[cfg(test)]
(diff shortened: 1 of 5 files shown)


mp4parse/src/lib.rs
```diff
@@ -35,7 +35,7 @@ use std::io::{Read, Take};
 mod macros;
 
 mod boxes;
-use boxes::{BoxType, FourCC};
+use crate::boxes::{BoxType, FourCC};
```
baumanj, line 38: This seems to work fine without the change. Is adding `crate::` a stylistic preference?

baumanj (approved):
Thanks for doing this!

mp4parse/src/lib.rs
kornelski: That's what `cargo fix` does. I presume it's for consistency, because `crate::` works anywhere, and without it it only works in the same file.

Issue #354, opened 2021-12-04
Implement `ItemDataBox` (`idat`)


baumanj:
This tends to be used for `grid` based items, so will be a prerequisite for https://github.com/mozilla/mp4parse-rust/issues/198

See ISOBMFF (ISO 14496-12:2020) § 8.11.11

PR #367, opened 2022-01-22
`ItemDataBox` support

baumanj:
Resolves https://github.com/mozilla/mp4parse-rust/issues/354
Resolves https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=42772

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -1189,10 +1189,10 @@ pub struct MediaContext {
 
 /// An ISOBMFF item as described by an iloc box. For the sake of avoiding copies,
 /// this can either be represented by the `Location` variant, which indicates
-/// where the data exists within a `MediaDataBox` stored separately, or the
-/// `Data` variant which owns the data. Unfortunately, it's not simple to
-/// represent this as a [`std::borrow::Cow`], or other reference-based type, because
-/// multiple instances may references different parts of the same [`MediaDataBox`]
+/// where the data exists within a `DataBox` stored separately, or the `Data`
+/// variant which owns the data. Unfortunately, it's not simple to represent
+/// this as a [`std::borrow::Cow`], or other reference-based type, because
+/// multiple instances may references different parts of the same [`DataBox`]
 /// and we want to avoid the copy that splitting the storage would entail.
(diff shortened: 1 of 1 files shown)


mp4parse/src/lib.rs
```diff
@@ -1419,10 +1424,15 @@ impl AvifContext {
...
-                    "IsobmffItem::Location requires the location exists in AvifContext::media_storage"
+                    "IsobmffItem::MdatLocation requires the location exists in AvifContext::media_storage"
                 );
             }
-            IsobmffItem::IdatLocation(_) => unimplemented!(),
+            IsobmffItem::IdatLocation(extent) => {
+                self.item_data_box
+                    .as_ref()
+                    .and_then(|idat| idat.get(extent))
+                    .unwrap_or_else(|| unreachable!("IsobmffItem::IdatLocation equires the location exists in AvifContext::item_data_box"))
```
kinetiknz, line 1434: Typo in "requires".

nit: I wonder if these uses of `unreachable!` should use `panic!` instead, since they're potentially parsing errors rather than code paths we never expect to hit.

kinetiknz (approved):
LGTM, thanks!

mp4parse/src/lib.rs
baumanj: I definitely agree with you that it _should_ be `panic!` if there's any way to reach this code, even with bad input. My assertion here (and in the above, mdat case) by using `unreachable!` is that there _isn't_ an input which will be able to reach that, and if it _is_ hit, that implies an error in the logic itself (and possibly bad input as well), not just bad input. In the case that I just haven't finished something, I use `unimplemented!`, but that's not the case here any longer.

In this specific case, I believe the only way there can be a `IsobmffItem::IdatLocation(extent)` value stored in `AvifItem::image_data`, is in the case that `AvifContext::item_data_box` has already been populated with a `DataBox` that covers `extent`. If my reasoning here is wrong, we should change it.

kinetiknz: Sorry, I should've looked more closely before making that suggestion - agree it's used appropriate in these situations.

PR #374, opened 2022-05-15
LayerSelector support

### LINKED ISSUE #373: Support `lsel` for decoding progressive image
https://github.com/AOMediaCodec/av1-avif/issues/180 changed standard(https://aomediacodec.github.io/av1-avif/#layer-selector-property) to require a `lsel` box with `layer_id` set to `0xFFFF` in an avif image to allow progressive decoding. Updated test images are yet to be merged: https://github.com/AOMediaCodec/av1-avif/pull/191.

With this change, mp4parse will reject any progressive image due to `lsel` being a mandatory unsupported box. We should at least be able to decode these images non-progressively.

I hope this can get into Firefox 102 ESR release, or we have to wait very long before being able to safely serve progressive AVIF images.

tongyuantongyu:
Resolves #373.


mp4parse/src/lib.rs
```diff
@@ -3785,6 +3808,22 @@ fn read_pixi<T: Read>(src: &mut BMFFBox<T>) -> Result<PixelInformation> {
     Ok(PixelInformation { bits_per_channel })
 }
 
+#[repr(C)]
+#[derive(Debug)]
+pub struct LayerSelector {
+    pub layer_id: u16
```
tongyuantongyu, line 3814: Test code need to access this value, so I made it public.

PR #378, opened 2022-09-01
Add some support for AVIF sequences.

Zaggy1024:
All properties not related to the `pitm` box have been moved from mp4parse_avif_get_image to mp4parse_avif_get_info. New fields for the color and alpha tracks are included in Mp4parseAvifInfo. To get the sample indices for those tracks, mp4parse_avif_get_indice_table was added as well.

These changes break C API backwards compatibility due to removing fields from Mp4parseAvifImage, so the version is bumped to 0.16.0.

I've implemented AVIF sequence decoding into Firefox using these changes, so they have been tested a bit at least. However, it's worth noting that I don't have access to the MIAF spec due to the price to buy it, so if anyone is able to confirm that anything related to that spec is sane that would be great.

I've also added the `pict` and `auxv` track handlers, which could potentially be fleshed out to verify validity, but I don't think that doing so would provide a tangible benefit to end users since it seems like libavif just uses the logic I've implemented for finding color and alpha tracks here.

I would imagine some new tests for these changes would be in order, but I'm unfortunately not familiar with cargo yet, so if those are desired, any pointers would to documentation would be helpful.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -299,19 +299,14 @@ impl Feature {
         match self {
             Self::Auxc
             | Self::Av1c
+            | Self::Avis
             | Self::Colr
             | Self::Imir
             | Self::Irot
             | Self::Ispe
             | Self::Pasp
             | Self::Pixi => true,
-            Self::A1lx
-            | Self::A1op
-            | Self::Clap
-            | Self::Grid
-            | Self::Ipro
-            | Self::Lsel
-            | Self::Avis => false,
+            Self::A1lx | Self::A1op | Self::Clap | Self::Grid | Self::Ipro | Self::Lsel => false,
         }
     }
 }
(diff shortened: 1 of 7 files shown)


ChunMinChang:
It seems these will be imported to gecko in [BMO 1788119](https://bugzilla.mozilla.org/show_bug.cgi?id=1788119)

Zaggy1024:
Bumping this since I've updated the tests that are relevant to AVIS support to pass with these changes.

Let me know if there are any new tests I should add as well, I'd be happy to add more coverage.

mp4parse/src/lib.rs
```diff
@@ -4444,6 +4483,32 @@ fn read_mdia<T: Read>(f: &mut BMFFBox<T>, track: &mut Track) -> Result<()> {
...
+            }
+            _ => skip_box_content(&mut b)?,
+        };
+        check_parser_state!(b.content);
+    }
+    Ok(TrackReferenceBox { references })
+}
+
+fn read_tref_auxl<T: Read>(f: &mut BMFFBox<T>) -> Result<TrackReference> {
+    let num_track_ids = (f.bytes_left() / 4).try_into()?;
```
kinetiknz, line 4503: nit: Make this `f.bytes_left() / size_of::<u32>()` for clarity.

mp4parse_capi/src/lib.rs
```diff
@@ -1108,26 +1262,62 @@ pub unsafe extern "C" fn mp4parse_get_indice_table(
...
+///
+/// This function is unsafe because it dereferences both the parser and
+/// indices raw pointers passed to it. Callers should ensure the parser
+/// points to a valid `Mp4parseAvifParser` and indices points to a valid
+/// `Mp4parseByteData`.
+#[no_mangle]
+pub unsafe extern "C" fn mp4parse_avif_get_indice_table(
+    parser: *mut Mp4parseAvifParser,
+    track_id: u32,
+    indices: &mut Mp4parseByteData,
```
kinetiknz, line 1286: Use `*mut Mp4parseByteData` here and null-check before use.

kinetiknz (approved):
Apologies for not reviewing this in a timely manner, and thanks for working on this!

Zaggy1024:
Thanks for the review! I've fixed the issues you mentioned, and squashed the "Moved pitm" commit into the first commit since it was mainly a bit of a fixup of that commit.

Should be good to go if you don't see any more issues!

Zaggy1024:
Looks like I missed a couple formatting issues. I fixed those, ran cargo fmt locally and force-pushed again, so hopefully it should be all good now.

Zaggy1024:
Sorry for the multiple CI attempts, this is my first time working with cargo so I wasn't aware of clippy. Hopefully I've fixed it so that it passes linting now.

kinetiknz:
No problem, thanks for the quick follow-ups.  Don't worry about the nightly-only clippy failures, I'll sort them out separately.

PR #383, opened 2022-12-08
Couple of minor fixes

rillian:
I don't think this is enough to get CI passing again, but it's a few positive steps...

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,20 @@
+name: Security audit
+on:
+  push:
+    paths: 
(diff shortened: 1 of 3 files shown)


rillian:
Looks like `actions-rs` needs blessing in the repo allow-list for the `cargo audit` check.

kinetiknz:
Thanks! I merged the criterion update.  The clippy fix already landed via another patch.

It looks like adding new action sources to the org's allow-list is not straightforward, mainly for security reasons, so I've left that change out.  It'd be great to run an audit in CI, perhaps we can look at taking a different approach.

PR #385, opened 2023-01-16
Read the loop count from AVIS files.

Zaggy1024:
The loop count is calculated as the number of times the animation must play to fill the `trak` duration based on the actual animation duration stored in the `elst` box.

I'm unsure if it will handle edit lists with multiple edits correctly, since it'll only take the first duration it finds, from my understanding of the `elst` parsing function, but I'm sure it's very unlikely to see more than one edit in practice.


Zaggy1024:
For reference, this comment is useful for the specifics of how this should work:

https://github.com/AOMediaCodec/av1-avif/issues/73#issuecomment-605335922

kinetiknz (approved):
LGTM, thanks!

It'd be good to add a test for the case where the loop count is not an integer and clarify the expected handling around that at the mp4parse C API level.

Zaggy1024:
> It'd be good to add a test for the case where the loop count is not an integer and clarify the expected handling around that at the mp4parse C API level.

Done! Lemme know if there's anything else, otherwise I think it should be good to go.

kinetiknz:
Thanks!

PR #388, opened 2023-01-25
Support unknown (zero) sized `mdat` boxes in AVIF.

kinetiknz:
Add basic support for unknown (zero) sized `mdat` boxes in AVIF.  Ideally, the `BMFFBox` `contents` would be limited to the remaining size of the file and everything would just work, but the file length is not available via the `Mp4ParseIo` interface (and may not be query-able for some streams) so it's not possible to rely on that.  Instead, when encountering a zero sized `mdat`, read the remaining file in chunks until end of file.

This partially addresses the issue reported in [BMO 1805436](https://bugzilla.mozilla.org/show_bug.cgi?id=1805436).


ashleyz (approved):
LGTM!

PR #389, opened 2023-01-26
Remove error for when an AVIF has no `iloc` box.

Zaggy1024:
Item locations can be omitted for AVIS-major files which don't contain `pitm` boxes. When no `iloc` is present, the locations map is default-initialized, and if the `pitm` references a nonexistent item, an error is thrown in strict mode.

I've added a couple test cases that work in Chrome but not in the upcoming Firefox AVIS patch. This should enable it to play those animations as expected.


kinetiknz:
LGTM, thanks!

PR #391, opened 2023-01-28
Read the looping flag when `elst` box is empty.

Zaggy1024:
The test file with no `pitm` and no `iloc` boxes should have had a loop count of 0 due to their `elst` box flags being set to 0, but the looping field that was read was being skipped due to the empty edit list.

Note: these test files had to be modified manually using MP4Box in gpac, libavif will not write files with no edits.


Zaggy1024:
Added a fix for the clippy errors that were preventing CI, let me know if you'd prefer that the commit is in a separate PR instead.

kinetiknz (approved):
Thanks!

PR #392, opened 2023-02-04
Don't panic when a box's size is smaller than its header.

Zaggy1024:
Fixes [Firefox Bug #1814736](https://bugzilla.mozilla.org/show_bug.cgi?id=1814736).


kinetiknz (approved):
This looks like fallout from https://github.com/mozilla/mp4parse-rust/pull/388.  Thanks for fixing this!

PR #393, opened 2023-02-04
Allow skipped top-level boxes to extend past EOF.

Zaggy1024:
This allows files with garbage appended to the end to still load.

Fixes [Firefox Bug #1814075](https://bugzilla.mozilla.org/show_bug.cgi?id=1814075).

This behavior isn't necessarily ideal, but other checks should fail if the file is unable to be parsed/decoded. The file in question as well as the test added here load successfully in Chrome,


mp4parse/src/lib.rs
```diff
@@ -2564,7 +2564,13 @@ pub fn read_avif<T: Read>(f: &mut T, strictness: ParseStrictness) -> Result<Avif
                 };
                 media_storage.push(DataBox::from_mdat(file_offset, data))?;
             }
-            _ => skip_box_content(&mut b)?,
+            _ => {
+                skip_box_content(&mut b)?;
+                // Allow garbage at EOF if we aren't in strict mode.
+                if b.bytes_left() > 0 && strictness != ParseStrictness::Strict {
```
kinetiknz, line 2570: Thanks for the fix!

I think allowing (ignoring) trailing unknown format data generally is going to require deeper changes in the parser.  Your fix works in this case since the trailing data in `contact.avif` happens to look enough like a BMFF box, but unknown data could trigger errors or panics in the `iter.next_box()` code path.

In terms of the code change here, the existing code is skipping unknown BMFF boxes anywhere at the top level of the AVIF parse, but I'd prefer ignoring unknown format data only at the EOF.  That might require adding some state to `read_avif` to track if the parse is complete with respect to the boxes we require (if possible).

Do you have any additional details on how Chrome's AVIF parser handles this situation?

kinetiknz (commented):
Thanks for the fix.  I'm fine with merging this as-is to solve the immediate issue if needed, but additional changes may be required to ignore trailing unknown format data.

Zaggy1024:
Replied in the review comment, but to summarize, after I investigated libavif, I'm satisfied this solution _should_ be fine if you have anything to add. I'll make it skip errors for the `skip_box_content()` call while at EOF before merging though.

mp4parse/src/lib.rs
Zaggy1024: > I think allowing (ignoring) trailing unknown format data generally is going to require deeper changes in the parser. Your fix works in this case since the trailing data in `contact.avif` happens to look enough like a BMFF box, but unknown data could trigger errors or panics in the `iter.next_box()` code path.

I may be wrong, but I don't think I see any places where `skip_box_content()` or `iter.next_box()` would panic, unless it can happen in the read functions provided by the reader or the read callback passed to `mp4parse_avif_new`. Any panics that happen in `skip_box_content` also seem like they would be present with or without this patch.

This did make me take a closer look at `read_box_header()`, though, and I realized that those field reads will probably error, so I need to also silently absorb the error from `skip_box_content()` here if this will be our solution.

> In terms of the code change here, the existing code is skipping unknown BMFF boxes anywhere at the top level of the AVIF parse, but I'd prefer ignoring unknown format data only at the EOF. That might require adding some state to `read_avif` to track if the parse is complete with respect to the boxes we require (if possible).

The `b.bytes_left() > 0` check makes sure that we're at EOF. Do you mean that you want `skip_box_content()` to only be valid at the end of the box iterator? Is there some requirement in the spec that the ftyp, meta, etc boxes in a particular order before any other boxes? I had assumed that the spec left room for new boxes to be added to the top level if necessary which was why we skip any unknowns here.

> Do you have any additional details on how Chrome's AVIF parser handles this situation?

It looks like it just [exits parsing](https://github.com/AOMediaCodec/libavif/blob/a8d6d1b098f08677e5a0c587e1bfb88d552b656f/src/read.c#L3047-L3049) as soon as it has the data it needs based on the `ftyp` brands, which means it's a bit more permissive than even this patch makes mp4parse. Given that nobody has yet complained about files with duplicate boxes not loading (I assume), I think that may be unnecessary, at least until progressive decode is a thing. (Haven't even given that much thought yet, but as I understand it mp4parse has no facilities for parsing just a partially-loaded file?)

Zaggy1024: > > I think allowing (ignoring) trailing unknown format data generally is going to require deeper changes in the parser. Your fix works in this case since the trailing data in `contact.avif` happens to look enough like a BMFF box, but unknown data could trigger errors or panics in the `iter.next_box()` code path.
> 
> I may be wrong, but I don't think I see any places where `skip_box_content()` or `iter.next_box()` would panic, unless it can happen in the read functions provided by the reader or the read callback passed to `mp4parse_avif_new`. Any panics that happen in `skip_box_content` also seem like they would be present with or without this patch.

Actually, I forgot that iter.next_box() error wouldn't be included in the error check in the match statement, my bad. I can probably throw together something to ignore those errors as well.

Zaggy1024:
@kinetiknz I've updated it with some new behavior that should be a lot more consistent at EOF. I made another test case and realized that `read_box_header()` had no way to signal that it ran out of data midway through parsing a box. This meant that the parser was already silently completing parsing on boxes that are smaller than their header requires. I've changed it to return an `Option` so that it can signal that, which allows me to detect EOF in the top-level parse and handle it based on strictness.

Hopefully that covers all your concerns, let me know if you need me to change anything!

mp4parse/src/lib.rs
kinetiknz: > I may be wrong, but I don't think I see any places where `skip_box_content()` or `iter.next_box()` would panic, unless it can happen in the read functions provided by the reader or the read callback passed to `mp4parse_avif_new`. Any panics that happen in `skip_box_content` also seem like they would be present with or without this patch.

I was thinking of the assert in `read_box_header` and the size/offset handling in `skip_box_content` (although the latter is not an issue since https://github.com/mozilla/mp4parse-rust/pull/392).   It may be safe to remove the assert, but it'd pay to review any other code using the raw box size/offset values.

> It looks like it just [exits parsing](https://github.com/AOMediaCodec/libavif/blob/a8d6d1b098f08677e5a0c587e1bfb88d552b656f/src/read.c#L3047-L3049) as soon as it has the data it needs based on the `ftyp` brands, which means it's a bit more permissive than even this patch makes mp4parse. Given that nobody has yet complained about files with duplicate boxes not loading (I assume), I think that may be unnecessary, at least until progressive decode is a thing. (Haven't even given that much thought yet, but as I understand it mp4parse has no facilities for parsing just a partially-loaded file?)

Thanks for the details - I feel that solution is cleaner, but let's try making this approach work first for simplicity.

kinetiknz (approved):
Thanks, let's give this a shot!

mp4parse/src/lib.rs
Zaggy1024: > I was thinking of the assert in `read_box_header` and the size/offset handling in `skip_box_content` (although the latter is not an issue since #392). It may be safe to remove the assert, but it'd pay to review any other code using the raw box size/offset values.

Ah, I see. I think it would probably be worth changing that assert to an error instead so that it's non-fatal, it looks like bad data could have caused a crash there even before, want me to make that change in this PR quick?

> Thanks for the details - I feel that solution is cleaner, but let's try making this approach work first for simplicity.

Yeah, as long as we're not trying to verify the file it definitely makes sense to do things more lazily. Once I have a better understanding of progressive decoding for AVIF I might think about how this fits into that.

kinetiknz: Sounds good, thanks!

Issue #395, opened 2023-02-09
Badly formed files can cause `create_sample_table()` to OOM.


Zaggy1024:
See Gecko [Bug 1661368](https://bugzilla.mozilla.org/show_bug.cgi?id=1661368) and more recently [Bug 1814791](https://bugzilla.mozilla.org/show_bug.cgi?id=1814791).

There is currently no validation of whether a file with an `stsc` box with a large sample count is actually referencing offsets that fall within the file, which allows `create_sample_table()` to allocate an arbitrarily large vector and OOM.

We could just set a hard limit on the number of samples it can allocate, but I don't know if that's the best solution.

We can also fail early when the relevant sample boxes have mismatched sample counts, but I'm not sure if there are files that currently parse and break that rule, and as tnikkel pointed out, the fuzzer may still find a way to create a matching sample count that causes OOM. Might still be worth implementing this if it makes sense to fail that way.

I think the most ideal solution is instead to make `create_sample_size()` aware of the amount of data in the file that is currently available to the caller, and use the `stsz` box to determine how many samples the vector needs to allocate to reach EOF. For cases where the full size is not known beforehand, API users would have to call `get_indices` multiple times and be aware that their pointers will be invalid, but that doesn't seem like a difficult problem.

What I wonder is whether the available data is something that should be accessible from other parts of mp4parse as well, for example to allow the box parsing to indicate that it hit the end of the available data but can continue parsing when more data is available.

Feel free to let me know if this idea doesn't make sense, we can use this issue to brainstorm more solutions if necessary.

Zaggy1024:
Looks like #394 is semi-related, and the limit it uses could also apply to `create_sample_table()`. However, since we're not actually reading from the offsets we create in the table, it won't actually exit when an invalid offset is reached, so it would only make the file size-based limit easier to implement.

I'm throwing up a proof of concept for that file size check now in a WIP pull request.

Zaggy1024:
Perhaps a less complicated way to prevent this issue is for the API to provide the table in batches, which seems like it may be a better general solution. However, it would require more Gecko changes to get that working, since the MP4 demuxer currently just assumes that the entire indice is immediately available.

kinetiknz:
[...]

PR #396, opened 2023-02-10
Limit the allocated size of the sample indice table by the file size.

Zaggy1024:
This is a proof of concept for my solution to slow OOMs in `create_sample_table()` laid out in #395.


Zaggy1024:
If this issue is able to occur on the MP4 side, this currently doesn't affect that, since the MP4 code in Gecko seems more complicated to do this with. If I can figure out how to get the size of the downloaded data there, I can probably modify it to request indices as the available data increases.

PR #399, opened 2023-03-13
Expose time in ticks relative to a timebase and not in microseconds

padenot:
As discussed.

### DIFF
--- a/mp4parse_capi/src/lib.rs
+++ b/mp4parse_capi/src/lib.rs
@@ -42,9 +42,7 @@ use std::io::Read;
 
 // Symbols we need from our rust api.
 use mp4parse::serialize_opus_header;
-use mp4parse::unstable::{
-    create_sample_table, media_time_to_us, track_time_to_us, CheckedInteger, Indice, Microseconds,
-};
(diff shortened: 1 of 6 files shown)


mp4parse_capi/src/lib.rs
```diff
@@ -631,20 +627,12 @@ pub unsafe extern "C" fn mp4parse_get_track_info(
...
 
-    if let (Some(track_timescale), Some(context_timescale)) = (track.timescale, context.timescale) {
-        let media_time: CheckedInteger<_> = match track
-            .media_time
-            .map_or(Some(Microseconds(0)), |media_time| {
-                track_time_to_us(media_time, track_timescale)
-            }) {
-            Some(time) => time.0.into(),
+    if let (Some(_), Some(_)) = (track.timescale, context.timescale) {
+        let media_time: CheckedInteger<u64> = match track.media_time {
```
kinetiknz, line 631: Need to carry over the `map_or(Some(0))` behaviour here and below for `empty_duration`.

kinetiknz (changes requested):
LGTM, thanks.  There are some test failures we'll need to fix before merging though.

kinetiknz (approved):

PR #407, opened 2023-07-10
empty_duration is in the mvhd's time base

padenot:
Regression from 888ce901bb604d9e3851f7e37e0dd66759d4a6b3.

@kinetiknz, what's best to add a test here, I'm not entirely sure on how to test with without bloating the repo with a big mp4.

This fixes the bug reported as [BMO#1842375](https://bugzilla.mozilla.org/show_bug.cgi?id=1842375).

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -1888,135 +1888,6 @@ impl DataBox {
     }
 }
 
-#[cfg(test)]
-mod media_data_box_tests {
-    use super::*;
-
-    impl DataBox {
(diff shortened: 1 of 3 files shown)


kinetiknz (approved):
LGTM, thanks

We're somewhat relying on Gecko tests for the C API side of this, so there's not many tests calling mp4parse_get_track_info aside from what's in mp4parse_capi/tests.

padenot:
merging as-is, will fix the nightly-only clippy errors separately.

PR #408, opened 2023-07-14
Add HEVC support.

doublech:
Add H.265 support by identifying 'hev1' or 'hvc1' entry and also extracting extradata from 'hvcC' box. The logic is very similar to H.264.


fabricedesre:
Can a maintainer let us know if you are interested in that addition? If not we'll move to a fork but we'd prefer not to.
Thanks!

cc @kinetiknz @padenot

kinetiknz:
Apologies for the delay, I'll review this tomorrow.

alastor0325:
Hmm I didn't notice that we already had this PR for HEVC, so I also made another one in https://github.com/mozilla/mp4parse-rust/pull/410. @kinetiknz maybe you can review them to see if we should combine them together, or just use one of them. Thanks!

kinetiknz:
Sorry about the confusion - I've reviewed both PRs and have merged the result now.  I used PR 410 as the base since this will be used in Gecko, so we don't need to hide it behind a feature flag.

PR #410, opened 2023-08-05
Support HEVC.

alastor0325:
In order to support DRM HEVC via the media engine, we need to be able to parse HEVC.


kinetiknz:
Thanks! I've merged a couple of small changes from PR 408.

PR #413, opened 2023-08-23
Fix broken build on MacOS

SingingTree:
The macos-10.15 target has been deprecated and removed from github, per https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/

This updates the build target so to instead tartget macos-12. I've opted to target an explicit version as that's what's done for the other OS targets, rather than using `latest`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


kinetiknz:
Thanks! 👍

PR #416, opened 2024-01-10
Ignore unused trailing data in `mvhd`.

kinetiknz:
This addresses [BMO 1867424](https://bugzilla.mozilla.org/show_bug.cgi?id=1867424), where the linked media contains an unusual `mvhd` with 2 bytes of trailing data.  Since `mvhd` has a fixed set of fields, we were expecting the `mvhd` size to match the spec based on the size of the known fields (108 bytes for this v0 file, v1 would be larger due to 64-bit fields) but the `mvhd` box size is 110 bytes in this file.  To address this, we now skip the remaining unknown bytes based on the box's reported size.


ashleyz (approved):
LGTM!

padenot:
We need to land this.

PR #421, opened 2024-11-21
Update fallible_collections

torokati44:
So that this crate doesn't (transitively) depend on the aging hashbrown 0.13 series.


kinetiknz:
Thanks!  I'll merge #422 instead since it updates both copies, but I appreciate the patch.

PR #422, opened 2024-11-26
Update fallible_collections from 0.4 to 0.5

musicinmybrain:
This is a near-duplicate of https://github.com/mozilla/mp4parse-rust/pull/421, but it updates *both* `fallible_collections` dependencies.

I confirmed that `cargo test --workspace` still succeeds.


kinetiknz:
Thanks!

PR #427, opened 2025-04-16
Make BoxType, BoxHeader, BMFFBox, BoxIter, and read_moov pub

torokati44:
Oftentimes, it would be nice to be able to start playing an MP4 file before it's available in its entirety, e.g. because it's streamed over a network connection.

In these cases, `read_mp4` fails, because it always tries to read every box it finds as far as I could tell.

However, since `moov` is the main info box that contains track info, etc, at the minimum this has to be parsed before attempting to decode samples (frames) in a streaming fashion.

There even is a hint in a comment in the code about this: `"found moov {}, could stop pure 'moov' parser now"`

Making these items public facilitates such a use case.

Something like this will be necessary for https://github.com/ruffle-rs/ruffle/pull/14655.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -35,7 +35,7 @@ use std::io::{Read, Take};
 mod macros;
 
 mod boxes;
-use crate::boxes::{BoxType, FourCC};
+pub use crate::boxes::{BoxType, FourCC};
 
 // Unit tests.
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


kinetiknz (approved):
Thanks, seems reasonable.

kinetiknz:
Landed via 761f8d6857f06f396c587fa8500b4b6f39e1c560

PR #429, opened 2025-04-17
Also allow padding at the end of ctts boxes

torokati44:
This was necessary to successfully parse the following file:
https://www.mea-culpa.at/02_videos/gesch_175_580x432px_12fps_300kbit.f4v

It has 7080 bytes of extra stuff in its `ctts` box after the given number of sample counts and offsets.

Other media players and demuxers (like VLC or MP4Box.js) had no problem with this particular file.

I copied the call and comment from other similar box reader functions nearby.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -4760,7 +4760,8 @@ fn read_ctts<T: Read>(src: &mut BMFFBox<T>) -> Result<CompositionOffsetBox> {
         })?;
     }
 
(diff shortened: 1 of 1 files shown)


kinetiknz:
Thanks! This seems reasonable, but I think it'd be good to only allow this when the parser is not in strict mode.  I've adjusted your change to account for that and landed it in 49fc621b2f7c405fe520034c9911099a434ffa47.

PR #437, opened 2025-08-19
Use `edited_duration` in C API

### LINKED ISSUE #404: Get correct media duration for files produced using `afconvert`
https://user-images.githubusercontent.com/270485/236244141-be70cc83-e36c-4ff6-ac30-49d51227b339.mp4

Here's a file that is supposed to be exactly 0.5s long.

It has been produced by taking a wav file that's 0.5s long exactly and running this on a mac:

> afconvert -s 3 -f mp4f -d aac half-a-second-44110-1ch.wav

It's got an `elst` saying the file has a start time of 2112 (typical for macOS AAC encoder), and a segment duration of 22050 (which is correct).

`mp4parse-rust` reports a duration of 24576, because it uses the duration of the track. ` mp4parse_get_fragment_info` doesn't work because it doesn't look into the right box.

mmaker-gh:
Closes #404

I looked into this because it appeared, as of a few months ago, Twitter started to encode videos in such a way that made this defect pretty obvious where a random gap of a few seconds exists at the end of every video. [Example](https://video.twimg.com/amplify_video/1957525448004411397/vid/avc1/1280x720/PSLG1doRn2lEdpJO.mp4?tag=21)

### DIFF
--- a/mp4parse_capi/src/lib.rs
+++ b/mp4parse_capi/src/lib.rs
@@ -650,12 +650,24 @@ pub unsafe extern "C" fn mp4parse_get_track_info(
             None => return Mp4parseStatus::Invalid,
         };
 
-        match track.duration {
-            Some(duration) => info.duration = duration.0,
-            None => {
-                // Duration unknown; stagefright returns 0 for this.
-                info.duration = 0
+        // Use edited duration if available
+        info.duration = if let Some(edited_duration) = track.edited_duration {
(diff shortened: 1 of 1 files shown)


mmaker-gh:
Implemented in https://github.com/mozilla/mp4parse-rust/commit/eaf515df080310849aca8d86933278941e8b36ba instead. Closing.

PR #439, opened 2025-09-24
If an elst is present, use that to report the duration

### LINKED ISSUE #404: Get correct media duration for files produced using `afconvert`

padenot:
This fixes #404.


padenot:
This is a fix only in C API since the Rust API has the right numbers, but maybe we can do something in the rust layer.

PR #446, opened 2026-04-09
Add new wrappers for image-rs

### LINKED ISSUE #441: Allow reading Exif metadata from AVIF images
The [`image`](https://crates.io/crates/image) crate uses `mp4parse` to decode AVIF images. I'd like to implement Exif metadata extraction from AVIF there, but `mp4parse` does not expose a function to retrieve the Exif metadata in `AvifContext`.

There's already Rust code for writing Exif into AVIF files which can be used as a reference:

https://github.com/kornelski/avif-serialize/blob/f4f4f96e28d0bd8da9e085ec90220708cc49bbbd/src/lib.rs#L214-L231

Adding such a function should be quite straightforward for someone already familiar with the format. I'm not familiar with it at all, which is why I am opening an issue and not a PR.

amyspark:
Hi all,

This PR adds support for a few BMFF boxes that are required to speed up the AVIF decoder's setup in image-rs, avoiding having to decode the full image to query e.g. dimensions, bit depth, etc.

While at it, I've also added support for detecting and exposing the Exif and XMP metadata boxes.

All feedback is appreciated.

Fixes #441
Fixes #444

cc @Shnatsel

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -227,6 +227,8 @@ pub enum Status {
     IlocOffsetOverflow,
     ImageItemType,
     InfeFlagsNonzero,
+    InfeStringNoNul,
+    InfeStringNotUtf8,
     InvalidUtf8,
     IpcoIndexOverflow,
     IpmaBadIndex,
@@ -610,6 +612,14 @@ impl From<Status> for &str {
             Status::InfeFlagsNonzero => {
                 "'infe' flags field shall be 0 \
                  per ISOBMFF (ISO 14496-12:2020) § 8.11.6.2"
+            }
+            Status::InfeStringNoNul => {
+                "'infe' strings shall be null-terminated \
+                 per ISOBMFF (ISO 14496-12:2020) § 8.11.6.3"
+            }
+            Status::InfeStringNotUtf8 => {
+                "'infe' strings field shall be valid utf8 \
+                 per ISOBMFF (ISO 14496-12:2020) § 8.11.6.3"
             }
             Status::InvalidUtf8 => {
                 "invalid utf8"
@@ -1594,6 +1604,10 @@ pub struct AvifContext {
     pub sequence: Option<MediaContext>,
     /// A collection of unsupported features encountered during the parse
     pub unsupported_features: UnsupportedFeatures,
+    /// AVIF box containing the Exif metadata, if any
+    exif_metadata: Option<AvifItem>,
+    /// AVIF box containing the XMP metadata, if any
+    xmp_metadata: Option<AvifItem>,
 }
 
 impl AvifContext {
@@ -1655,6 +1669,18 @@ impl AvifContext {
         }
     }
 
+
+    pub fn exif_metadata(&self) -> Option<&[u8]> {
+        self.exif_metadata
+            .as_ref()
+            .map(|item| self.item_as_slice(item))
+    }
+
+    pub fn xmp_metadata(&self) -> Option<&[u8]> {
+        self.xmp_metadata
+            .as_ref()
+            .map(|item| self.item_as_slice(item))
+    }
 
     pub fn spatial_extents(&self) -> Result<&ImageSpatialExtentsProperty> {
         if let Some(primary_item) = &self.primary_item {
@@ -1970,7 +1996,7 @@ impl DataBox {
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 struct PropertyIndex(u16);
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, PartialOrd)]
-struct ItemId(u32);
+pub struct ItemId(u32);
 
 impl ItemId {
     fn read(src: &mut impl ReadBytesExt, version: u8) -> Result<ItemId> {
@@ -1986,9 +2012,13 @@ impl ItemId {
 /// See ISOBMFF (ISO 14496-12:2020) § 8.11.6
 /// Only versions {2, 3} are supported
 #[derive(Debug)]
-struct ItemInfoEntry {
-    item_id: ItemId,
-    item_type: u32,
+pub struct ItemInfoEntry {
+    pub item_id: ItemId,
+    pub item_type: u32,
+    pub item_name: TryVec<u8>,
+    pub content_type: Option<TryVec<u8>>,
+    pub content_encoding: Option<TryVec<u8>>,
+    pub uri_type: Option<TryVec<u8>>,
 }
 
 /// See ISOBMFF (ISO 14496-12:2020) § 8.11.12
@@ -2577,24 +2607,17 @@ pub fn read_avif<T: Read>(f: &mut T, strictness: ParseStrictness) -> Result<Avif
     debug!("alpha_item_id: {alpha_item_id:?}");
     let mut primary_item = None;
     let mut alpha_item = None;
+    let mut items: TryHashMap<ItemId, AvifItem> = Default::default();
 
     // store data or record location of relevant items
     for (item_id, loc) in iloc_items {
-        let item = if Some(item_id) == primary_item_id {
-            &mut primary_item
-        } else if Some(item_id) == alpha_item_id {
-            &mut alpha_item
-        } else {
-            continue;
-        };
-
-        assert!(item.is_none());
+        let mut item: Option<AvifItem> = None;
 
         // If our item is spread over multiple extents, we'll need to copy it
         // into a contiguous buffer. Otherwise, we can just store the extent
         // and return a pointer into the mdat/idat later to avoid the copy.
         if loc.extents.len() > 1 {
-            *item = Some(AvifItem::with_inline_data(item_id))
+            item = Some(AvifItem::with_inline_data(item_id))
         }
 
         trace!(
@@ -2607,10 +2630,10 @@ pub fn read_avif<T: Read>(f: &mut T, strictness: ParseStrictness) -> Result<Avif
         // true if the extent is successfully added to the AvifItem
         let mut find_and_add_to_item = |extent: &Extent, dat: &DataBox| -> Result<bool> {
             if let Some(extent_slice) = dat.get(extent) {
-                match item {
+                match &mut item {
                     None => {
                         trace!("Using IsobmffItem::Location");
-                        *item = Some(AvifItem {
+                        item = Some(AvifItem {
                             id: item_id,
                             image_data: dat.location(extent),
                         });
@@ -2670,6 +2693,14 @@ pub fn read_avif<T: Read>(f: &mut T, strictness: ParseStrictness) -> Result<Avif
         }
 
         assert!(item.is_some());
+
+        if Some(item_id) == primary_item_id {
+            primary_item = item;
+        } else if Some(item_id) == alpha_item_id {
+            alpha_item = item;
+        } else {
+            items.insert(item_id, item.unwrap())?;
+        };
     }
 
     if (primary_item_id.is_some() && primary_item.is_none())
@@ -2773,6 +2804,21 @@ pub fn read_avif<T: Read>(f: &mut T, strictness: ParseStrictness) -> Result<Avif
     check_image_item(&mut primary_item)?;
     check_image_item(&mut alpha_item)?;
 
+    let mut exif_metadata = None;
+    let mut xmp_metadata = None;
+
+    item_infos.into_iter().for_each(|item| {
+        if &item.item_type.to_be_bytes() == b"Exif" {
+            exif_metadata = items.remove(&item.item_id);
+        } else if &item.item_type.to_be_bytes() == b"mime"
+            && item
+                .content_type
+                .is_some_and(|v| v.as_slice() == b"application/rdf+xml")
+        {
+            xmp_metadata = items.remove(&item.item_id);
+        }
+    });
+
     Ok(AvifContext {
         strictness,
         media_storage,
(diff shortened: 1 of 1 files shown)


mp4parse/src/lib.rs
```diff
@@ -1640,6 +1644,56 @@ impl AvifContext {
...
+    }
+
+    pub fn exif_metadata(&self) -> Option<&[u8]> {
+        self.exif_metadata
+            .as_ref()
+            .map(|item| self.item_as_slice(item))
+    }
+
+    pub fn xmp_metadata_is_present(&self) -> bool {
+        self.primary_item.is_some()
```
linkmauve: This is the same code as on line 1663, for `exif_metadata_is_present()`, shouldn’t that be changed to this?
```suggestion
        self.xmp_metadata.is_some()
```

mp4parse/src/lib.rs
```diff
@@ -1640,6 +1644,56 @@ impl AvifContext {
...
+                Some(other_property) => panic!("property key mismatch: {:?}", other_property),
+                None => Err(Error::from(Status::Av1cMissing)),
+            }
+        } else {
+            Err(Error::from(Status::PitmMissing))
+        }
+    }
+
+    pub fn exif_metadata_is_present(&self) -> bool {
+        self.primary_item.is_some()
```
linkmauve: Same here?
```suggestion
        self.exif_metadata.is_some()
```

linkmauve (changes requested):

mp4parse/src/lib.rs
```diff
@@ -2501,24 +2623,17 @@ pub fn read_avif<T: Read>(f: &mut T, strictness: ParseStrictness) -> Result<Avif
     debug!("alpha_item_id: {alpha_item_id:?}");
     let mut primary_item = None;
     let mut alpha_item = None;
+    let mut items: std::collections::HashMap<ItemId, AvifItem> = Default::default();
```
kinetiknz: We don't permit infallible data structure use.  Use TryHashMap.

mp4parse/src/lib.rs
```diff
@@ -2200,6 +2302,26 @@ impl<T> Drop for BMFFBox<'_, T> {
     }
 }
 
+impl<T: ReadBytesExt> BMFFBox<'_, T> {
+    fn read_string(&mut self, strictness: ParseStrictness) -> Result<TryVec<u8>> {
```
kinetiknz: Don't add this.  The string parsing for `infe` can be handled inline.

mp4parse/src/lib.rs
```diff
@@ -1640,6 +1644,56 @@ impl AvifContext {
...
+                Some(ItemProperty::AV1Config(av1c)) => Ok(av1c),
+                Some(other_property) => panic!("property key mismatch: {:?}", other_property),
+                None => Err(Error::from(Status::Av1cMissing)),
+            }
+        } else {
+            Err(Error::from(Status::PitmMissing))
+        }
+    }
+
+    pub fn exif_metadata_is_present(&self) -> bool {
```
kinetiknz: `exif_metadata_is_present()` seems unnecessary, callers can just use `exif_metadata().is_some()`.  Same with `xmp_metadata_is_present()`.

mp4parse/src/lib.rs
```diff
@@ -1640,6 +1644,56 @@ impl AvifContext {
...
+        self.xmp_metadata.is_some()
+    }
+
+    pub fn xmp_metadata(&self) -> Option<&[u8]> {
+        self.xmp_metadata
+            .as_ref()
+            .map(|item| self.item_as_slice(item))
+    }
+
+    pub fn spatial_extents(&self) -> Result<&ImageSpatialExtentsProperty> {
```
kinetiknz, line 1685: The duplication in these additions with the `*_ptr()` variants needs to be addressed.

mp4parse/src/lib.rs
```diff
@@ -2912,15 +3052,30 @@ fn read_infe<T: Read>(
...
     if item_protection_index != 0 {
         unsupported_features.insert(Feature::Ipro);
-        Ok(None)
-    } else {
-        Ok(Some(ItemInfoEntry { item_id, item_type }))
+        skip_box_remain(src)?;
+        return Ok(None);
     }
+
+    let item_name = src.read_string(strictness)?;
```
kinetiknz: Use the pattern in `read_hdlr` to read nul-terminated strings.  Use correct error types per that pattern, not the generic `InvalidUtf8` that `read_string` returned.

With `skip_box_remain` moved, this is now broken when parsing item_type == `uri` boxes.  Also broken if the optional `content_encoding` field is not present, since these changes assume it is.

mp4parse/src/lib.rs
```diff
@@ -2912,15 +3052,30 @@ fn read_infe<T: Read>(
...
+
+    let (content_type, content_encoding) = if item_name.as_slice() == b"mime" {
```
kinetiknz: Isn't this supposed to check `item_type`?

kinetiknz (changes requested):
I haven't reviewed this thoroughly yet, but I've included some initial feedback.

The `iloc` read loop is potentially very expensive after these changes.  It should remain the way it was, but extended to read the required items.

The `read_infe`/`read_string` changes don't work correctly for multiple reasons, so I don't believe this code has been tested.

We'll need automated tests added to cover the changes in the core parser.

mp4parse/src/lib.rs
amyspark: I am not sure how to address it, since using the existing `_ptr` APIs requires `unsafe` blocks:

```
error[E0133]: dereference of raw pointer is unsafe and requires unsafe function or block
  --> src\codecs\avif\decoder.rs:99:21
   |
99 |             height: (*dimensions).image_height,
   |                     ^^^^^^^^^^^^^ dereference of raw pointer
   |
   = note: raw pointers may be null, dangling or unaligned; they can violate aliasing rules and cause data races: all of these are undefined behavior
```

As per https://github.com/mozilla/mp4parse-rust/issues/444#issuecomment-4173406001 you've been relying on cbindgen to use this crate, so didn't want to modify them.

mp4parse/src/lib.rs
amyspark: Pointing to the `read_hdlr` comment to better address this.

mp4parse/src/lib.rs
amyspark: > Use the pattern in read_hdlr to read nul-terminated strings.

There are four different instances of null-terminated UTF-8 strings, I would like them to not copy-paste code for each one. The next iteration turns it into a standalone function with two new error code enums, let me know if that's OK.

Also took care of the `uri` boxes.

> if the optional `content_encoding` field is not present

In the AVIF files I have here with XMP metadata, the field is present but empty. I couldn't find in 14496:12-2020 how to detect missing-but-optional strings, could you confirm whether this is done by expecting the box to end upon read?

amyspark:
Hi, is there anything I can do to help this forward?

kinetiknz:
> Hi, is there anything I can do to help this forward?

I think it'd be a good idea to separate this into two PRs.  It's fairly easy to review and merge the "Add safe wrappers for AVIF metadata boxes" change, the only issue with that is the need to reduce the duplication between the safe and `*_ptr()` variants.  That can be done with a safe helper for 

"Read and expose Exif and XMP metadata in the AvifContext" needs tests for the `infe` parser changes (it's clear this code still has not been tested - the `application/rdf+xml` comparison fails because of untrimmed NULs, for example), the string reading should use `read_into_try_vec` to bound the read length (and the result can then be split on NULs into the appropriate fields).  The `uri_type` reading path is taken for every non-`mime` type, causing parser failures in Normal/Strict mode on valid files.

mp4parse/src/lib.rs
kinetiknz:
```Rust
fn primary_property<'a, T>(
        &'a self,
        box_type: BoxType,
        extract: impl Fn(&'a ItemProperty) -> Option<&'a T>,
    ) -> Result<Option<&'a T>> {
        let primary = self.primary_item.as_ref().ok_or(Error::from(Status::PitmMissing))?;
        match self.item_properties.get(primary.id, box_type)? {
            Some(prop) => Ok(Some(extract(prop).expect("property key mismatch"))),
            None => Ok(None),
        }
    }
```

The safe and _ptr variants can then be rewritten as:

```Rust
  pub fn image_mirror(&self) -> Result<&ImageMirror> {
      self.primary_property(BoxType::ImageMirror, |p| match p {
          ItemProperty::Mirroring(v) => Some(v),
          _ => None,
      })?
      .ok_or_else(|| Status::ImirMissing.into())
  }

  pub fn image_mirror_ptr(&self) -> Result<*const ImageMirror> {
      match self.image_mirror() {
          Ok(v) => Ok(v),
          Err(Error::InvalidData(Status::PitmMissing | Status::ImirMissing)) => Ok(std::ptr::null()),
          Err(e) => Err(e),
      }
  }
```

for example.  The `spatial_extents` change needs to preserve the `fail_with_status_if` check.

mp4parse/src/lib.rs
```diff
@@ -1661,6 +1718,21 @@ impl AvifContext {
         }
     }
 
+    pub fn colour_information(&self) -> Result<&ColourInformation> {
```
kinetiknz, line 1721: Don't add this as-is.  item_properties.get() errors with Status::IprpConflict when multiple colr boxes are associated with the primary item but both an nclx and an ICC colr on  one item is legal per HEIF (and common in real AVIFs that want to specify both display primaries and a profile).  Add safe variants of the existing nclx_colour_information_ptr and use icc_colour_information directly.

mp4parse/src/lib.rs
```diff
@@ -1722,6 +1794,21 @@ impl AvifContext {
...
 
+    pub fn image_mirror(&self) -> Result<&ImageMirror> {
+        if let Some(primary_item) = &self.primary_item {
+            match self
+                .item_properties
+                .get(primary_item.id, BoxType::ImageMirror)?
+            {
+                Some(ItemProperty::Mirroring(imir)) => Ok(imir),
+                Some(other_property) => panic!("property key mismatch: {:?}", other_property),
+                None => Err(Error::from(Status::ItemTypeMissing)),
```
kinetiknz, line 1805: Error type should be Status::ImirMissing.

mp4parse/src/lib.rs
```diff
@@ -1737,6 +1824,21 @@ impl AvifContext {
...
 
+    pub fn pixel_aspect_ratio(&self) -> Result<&PixelAspectRatio> {
+        if let Some(primary_item) = &self.primary_item {
+            match self
+                .item_properties
+                .get(primary_item.id, BoxType::PixelAspectRatioBox)?
+            {
+                Some(ItemProperty::PixelAspectRatio(pasp)) => Ok(pasp),
+                Some(other_property) => panic!("property key mismatch: {:?}", other_property),
+                None => Err(Error::from(Status::ItemTypeMissing)),
```
kinetiknz, line 1835: Error type should be Status::PaspMissing.

mp4parse/src/lib.rs
```diff
@@ -1894,7 +1996,7 @@ impl DataBox {
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 struct PropertyIndex(u16);
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, PartialOrd)]
-struct ItemId(u32);
+pub struct ItemId(u32);
```
kinetiknz, line 1999: Why was this and ItemInfoEntry made `pub`?

mp4parse/src/lib.rs
```diff
@@ -1910,9 +2012,13 @@ impl ItemId {
 /// See ISOBMFF (ISO 14496-12:2020) § 8.11.6
 /// Only versions {2, 3} are supported
 #[derive(Debug)]
-struct ItemInfoEntry {
-    item_id: ItemId,
-    item_type: u32,
+pub struct ItemInfoEntry {
+    pub item_id: ItemId,
+    pub item_type: u32,
```
kinetiknz, line 2017: This should be `FourCC`.

mp4parse/src/lib.rs
```diff
@@ -2501,24 +2607,17 @@ pub fn read_avif<T: Read>(f: &mut T, strictness: ParseStrictness) -> Result<Avif
...
-        let item = if Some(item_id) == primary_item_id {
-            &mut primary_item
-        } else if Some(item_id) == alpha_item_id {
-            &mut alpha_item
-        } else {
-            continue;
-        };
-
-        assert!(item.is_none());
+        let mut item: Option<AvifItem> = None;
```
kinetiknz, line 2614: Revert these changes entirely.  This can be done more efficiently with:

Before the iloc_items loop:
```Rust
  let exif_item_id = item_infos.iter()
      .find(|i| i.item_type.to_be_bytes() == *b"Exif")
      .map(|i| i.item_id);
  let xmp_item_id = item_infos.iter()
      .find(|i| i.item_type.to_be_bytes() == *b"mime"
          && i.content_type.as_ref()
              .is_some_and(|ct| ct.as_slice() == b"application/rdf+xml"))
      .map(|i| i.item_id);
```

Inside the loop:
```Rust
  let item = if Some(item_id) == primary_item_id {
      &mut primary_item
  } else if Some(item_id) == alpha_item_id {
      &mut alpha_item
  } else if Some(item_id) == exif_item_id {
      &mut exif_metadata
  } else if Some(item_id) == xmp_item_id {
      &mut xmp_metadata
  } else {
      continue;
  };
```

kinetiknz (changes requested):

PR #447, opened 2026-04-16
Parse colr box CICP data for video tracks

alastor0325:
## Summary

  The `read_colr` function and `ColourInformationBox` parsing already existed in this codebase, but were only wired up
   to the AVIF/image item path (`read_iprp`). The video track path (`read_video_sample_entry`) silently discarded the
  `colr` box via the default `skip_box_content` arm.

  This PR connects the existing `read_colr` to video tracks and exposes the CICP values through the C API.

  ## Changes

  **`mp4parse` (Layer 1 — parser)**
  - Add `colour_info: Option<ColourInformation>` field to `VideoSampleEntry`
  - Add `BoxType::ColourInformationBox` arm in `read_video_sample_entry` to call the existing `read_colr`
  - Add `strictness` parameter to `read_video_sample_entry` (needed to pass to `read_colr`, consistent with
  `read_audio_sample_entry`)
  - Make `NclxColourInformation` fields `pub` so `mp4parse_capi` can read them
  - Duplicate `colr` boxes violate ISO 14496-12:2020 §12.1.5 — we warn and keep the first rather than failing, to
  avoid breaking playback of malformed content

  **`mp4parse_capi` (Layer 2 — C API)**
  - Add four fields to `Mp4parseTrackVideoSampleInfo`:
    - `has_colour_info: bool` — canonical presence flag; needed because MC=0 is a valid CICP value (Identity/GBR), not
   an absence indicator
    - `colour_primaries: u8`
    - `transfer_characteristics: u8`
    - `matrix_coefficients: u8`
    - `full_range_flag: bool`
  - Populate these fields from `video.colour_info` in `mp4parse_get_track_video_info_safe`

  ## Tests

  Six integration tests with ffmpeg-generated MP4 files (`-movflags +write_colr`, `-x264-params
  colorprim/transfer/colormatrix`):

  | Test | cp | tc | mc | full_range |
  |---|---|---|---|---|
  | HDR10 | 9 | 16 | 9 | false |
  | HDR10 full-range | 9 | 16 | 9 | true |
  | HLG | 9 | 18 | 9 | false |
  | HLG full-range | 9 | 18 | 9 | true |
  | RGB/Identity (MC=0) | 1 | 1 | 0 | false |
  | No `colr` box | — | — | — | — |


alastor0325:
@kinetiknz could you help review this? This is a fix for bug 2030296 where we miss parse `colr` for mp4 playback. Thanks!

mp4parse/src/lib.rs
```diff
@@ -5683,6 +5688,19 @@ fn read_video_sample_entry<T: Read>(src: &mut BMFFBox<T>) -> Result<SampleEntry>
                 }
                 debug!("Parsed pasp box: {pasp:?}, PAR {pixel_aspect_ratio:?}");
             }
+            BoxType::ColourInformationBox => {
+                if colour_info.is_some() {
+                    // ISO 14496-12:2020 §12.1.5 requires at most one ColourInformationBox
+                    // per sample entry. Duplicates are a spec violation, but we tolerate
+                    // them to avoid breaking playback of malformed content.
+                    warn!("Multiple colr boxes in video sample entry, keeping first");
```
kinetiknz, line 5715: Can we add:

```
      fail_with_status_if(
          strictness != ParseStrictness::Permissive,
          Status::ColrBadQuantityBMFF,
      )?;
```

here?

Note that we should add a new ColrBadQuantityBMFF error since ColrBadQuantity has a HEIF-specific error message associated with it.

mp4parse/src/lib.rs
```diff
@@ -5683,6 +5688,19 @@ fn read_video_sample_entry<T: Read>(src: &mut BMFFBox<T>) -> Result<SampleEntry>
...
+                    skip_box_content(&mut b)?;
+                } else {
+                    let colr = read_colr(&mut b, strictness)?;
```
kinetiknz: I'm worried that `read_colr` is too strict for MP4 since it's only been used for AVIF so far.  I'm not sure how common it is to find `nclc` boxes in existing content, but before this change we'd ignore them and now we'll fail to parse the file due to `read_colr` being strict about supported types.

Something like changing the `read_colr` return type to:

```
enum ParsedColourInformation {
      Supported(ColourInformation),
      Unsupported(FourCC),
  }
```

Then [this path](https://github.com/mozilla/mp4parse-rust/blob/master/mp4parse/src/lib.rs#L3806) can `skip_box_content` and return `ParsedColourInformation::Unsupported(FourCC::from(colour_type))`.  The AVIF call path can convert `Unsupported` into `Status::ColrBadType` at the call site, and the MP4 call path can just log a warning.

kinetiknz (changes requested):
Thanks! Looks good, just a couple of things to address.

kinetiknz (approved):
LGTM, thanks!

PR #448, opened 2026-04-21
Parse mdcv and clli boxes in video sample entries

alastor0325:
### Background

`MasteringDisplayColourVolumeBox` (`mdcv`) and `ContentLightLevelBox` (`clli`) are defined in ISO/IEC 14496-12 as codec-agnostic child boxes of `VisualSampleEntry`. They carry HDR10 static metadata:

- `mdcv` — SMPTE ST 2086 mastering display colour volume (primaries, white point, luminance range)
- `clli` — CTA-861.3 content light level (MaxCLL, MaxFALL)

Because they sit on `VisualSampleEntry` rather than inside any codec-specific box, they apply to all video codecs: `hvc1`, `av01`, `vp09`, `avc1`, etc. This is particularly important for **VP9-in-MP4**, which has no in-bitstream HDR metadata mechanism and relies entirely on these container boxes.

These boxes are distinct from the older `SmDm`/`CoLL` boxes from the WebM VP9 binding spec. They carry the same data, but `mdcv`/`clli` are plain `Box` (no version/flags header), while `SmDm`/`CoLL` are `FullBox`. Chromium handles both; this PR adds support for the ISO-standardized `mdcv`/`clli` variant.

### Design note: raw integers vs floats

We store raw fixed-point integers rather than converting to floats at parse time. This is lossless and matches the wire format exactly, letting the caller decide how to interpret the values. Chromium’s [implementation](https://source.chromium.org/chromium/chromium/src/+/main:media/formats/mp4/box_definitions.h;l=307-316;drc=f42c791f42978cad13a35edfb7ca398b2e10ff48) converts to floats inside `Parse()`—both approaches are valid; we prefer keeping the library as a faithful parser.

### Testing

Unit tests added to `mp4parse/src/tests.rs`:

- `read_mdcv` — synthetic 24-byte payload, asserts all 8 fields round-trip correctly
- `read_clli` — synthetic 4-byte payload, asserts both light level fields


alastor0325:
This is for [2033628](https://bugzilla.mozilla.org/show_bug.cgi?id=2033628).

alastor0325:
@kinetiknz could you help me review this? Thanks!

mp4parse/src/boxes.rs
```diff
@@ -123,6 +123,8 @@ box_database!(
     ItemPropertyContainerBox          0x6970_636f, // "ipco"
     ItemPropertyAssociationBox        0x6970_6d61, // "ipma"
     ColourInformationBox              0x636f_6c72, // "colr"
+    MasteringDisplayColourVolumeBox   0x6d646376, // "mdcv"
+    ContentLightLevelBox              0x636c6c69, // "clli"
```
kinetiknz, line 127: nit: match formatting of existing code with a `_` separator

mp4parse/src/lib.rs
```diff
@@ -3701,6 +3729,32 @@ fn read_pasp<T: Read>(src: &mut BMFFBox<T>) -> Result<PixelAspectRatio> {
     })
 }
 
+/// Parse mastering display colour volume box (ISO 14496-12).
+fn read_mdcv<T: Read>(src: &mut BMFFBox<T>) -> Result<MasteringDisplayColourVolume> {
+    let display_primaries_x = [be_u16(src)?, be_u16(src)?, be_u16(src)?];
+    let display_primaries_y = [be_u16(src)?, be_u16(src)?, be_u16(src)?];
```
kinetiknz, line 3735: The read order looks wrong here with respect to the docs on `MasteringDisplayColourVolume` above.  As far as I can tell the order in the box seems to be: G_x, G_y, B_x, B_y, R_x, R_y.

kinetiknz (changes requested):

kinetiknz (approved):
LGTM, thanks!

PR #449, opened 2026-04-23
Fix esds DSI parsing for truncated and duplicate DecSpecificInfo.

kinetiknz:
Record AOT, sample rate and channel count before the GASpecificConfig bit reads that can hit EOF on a truncated DSI.  find_descriptor swallows BitReaderError in non-strict mode, so storing the fields afterwards left a playable-looking AAC track with no usable config and silent playback for HE-AAC streams with explicit SBR signalling.  This is a fix for https://bugzilla.mozilla.org/show_bug.cgi?id=2011644

Also keep the first DecSpecificInfo and ignore subsequent ones in non-strict mode, matching Chromium and ffmpeg.  The previous code concatenated DSI bytes and overwrote the parsed audio fields from the second DSI, leaving the ES_Descriptor inconsistent when the two disagreed.  Strict mode still rejects duplicates with EsdsDecSpecificInfoTagQuantity.

### DIFF
--- a/mp4parse/src/lib.rs
+++ b/mp4parse/src/lib.rs
@@ -5151,6 +5151,20 @@ fn read_ds_descriptor(
     esds: &mut ES_Descriptor,
     strictness: ParseStrictness,
 ) -> Result<()> {
+    // Keep the first DecSpecificInfo and ignore subsequent entries in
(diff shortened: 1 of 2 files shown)


kinetiknz:

alastor0325 (approved):
LGTM, thanks!

PR #450, opened 2026-06-04
Accept multiple colr boxes in a video sample entry

alastor0325:
### Summary

A video sample entry containing more than one `colr` (`ColourInformationBox`) is currently rejected with `ColrBadQuantityBMFF` at any strictness level other than `Permissive`. Since Firefox parses both MP4 video (`mp4parse_new`) and animated AVIF (`mp4parse_avif_new`) using `ParseStrictness::Normal`, this causes otherwise valid files to fail parsing entirely.

This breaks real-world content, most notably the **backward-compatible HLG** signaling convention, which emits two adjacent `nclx colr` boxes (an HLG box followed by a BT.2020 fallback), as produced by libavif and observed in the wild.

### Spec Basis

ISO/IEC 14496-12:2015 §12.1.5.1 ("Colour information" within `VisualSampleEntry`) states:

> "Colour information may be supplied in one or more ColourInformationBoxes placed in a VisualSampleEntry. These should be placed in order ... starting with the most accurate ... These are advisory and ... there is no normative behaviour associated with them; a reader may choose to use the most suitable. A ColourInformationBox with an unknown colour type may be ignored."

Multiple `colr` boxes are therefore explicitly permitted, and readers are free to select whichever box they consider most appropriate. The parser already preserves the first box's data; only the error handling was inconsistent with that behavior.

### Change

Demote duplicate-`colr` handling in `read_video_sample_entry` from a hard parse error under `Normal` strictness to a warn-and-keep-first behavior. The check now triggers only under `Strict`, preserving the ability for strict consumers to receive `ColrBadQuantityBMFF`.

The first box's CICP data continues to be surfaced unchanged. The AVIF/HEIF item-property path (`ColrBadQuantity`) is unaffected.

### Test

Adds `video_colr_nclx_two_colr` to `mp4parse_capi/tests/test_colour_info.rs` using a fixture containing two adjacent `nclx colr` boxes (HLG transfer=18 followed by transfer=14). The test verifies that parsing succeeds and that the first box (transfer=18) is surfaced.

Verified red before the change (`ColrBadQuantityBMFF`) and green after.

### Fixes

- https://bugzilla.mozilla.org/show_bug.cgi?id=2044320 (MP4 video playback)
- https://bugzilla.mozilla.org/show_bug.cgi?id=2044532 (animated AVIF)


alastor0325:
@kinetiknz could you help me review this? Thanks!

PR #451, opened 2026-06-10
Make duplicate colr detection per colour_type in video sample entries

kinetiknz:
### Summary

Follow-up to #450, which demoted duplicate-`colr` rejection in video sample entries to `Strict`-only. This cleans up the loose ends from that review:

1. **Per-`colour_type` duplicate detection.** ISOBMFF (ISO 14496-12:2020 § 12.1.5) permits multiple `colr` boxes in a `VisualSampleEntry`, at most one for a given `colour_type` — the `nclx` + ICC pairing is the intended use. Previously *any* second `colr` box triggered the warn/`Strict`-fail path, penalizing that valid combination. `read_video_sample_entry` now tracks the colour types seen and flags only a repeated `colour_type`, matching the per-type handling already used for HEIF item properties (`ColrBadQuantity`). `read_colr` now takes the already-read `colour_type` so the caller can make this decision without consuming the box.

2. **Fix the `ColrBadQuantityBMFF` description.** It still claimed "Each sample entry shall have at most one ColourInformationBox (colr)", contradicting the spec text cited at the check site since #450. It now describes the per-`colour_type` constraint actually enforced.

3. **Regression tests for the `Strict` path.** Nothing pinned that `Strict` still returns `ColrBadQuantityBMFF` (the capi tests only run at `Normal`). Adds two tests to `mp4parse/tests/public.rs`:
   - `public_video_two_nclx_colr_boxes`: the dual-`nclx` HLG fixture from #450 parses at `Permissive`/`Normal` (surfacing the first box, transfer=18) and fails with `ColrBadQuantityBMFF` at `Strict`.
   - `public_video_nclx_and_icc_colr_boxes`: an `nclx` + `rICC` variant of the same fixture parses at **all** strictness levels.

The first surfaced `colr` box continues to win; behavior at `Permissive` and `Normal` is unchanged for files #450 unblocked.

Not addressed here (pending evidence of real-world files): demoting the HEIF item-property `ColrBadQuantity` check, which still fails same-type duplicates at `Normal` for still-image AVIF.


kinetiknz:

alastor0325 (approved):
LGTM, thanks!

PR #452, opened 2026-06-13
Tolerate nclx colr boxes missing the full_range_flag byte

kinetiknz:
Some muxers write an 18-byte nclx colr box that omits the trailing full_range_flag/reserved byte, matching the layout of the QTFF 'nclc' colour type. read_colr failed on these with a fatal BitReaderError, rejecting the whole file, while ffmpeg, Chrome and Safari accept such boxes and treat the range as limited.

Default full_range_flag to false in this case and only reject the short box under ParseStrictness::Strict. Fixes bug 2047239.


kinetiknz:

alastor0325 (approved):
LGTM, thanks!

PR #456, opened 2026-09-16
Parse a1lx and expose item extents for layered AVIF images

jakearchibald:
This is part of https://bugzilla.mozilla.org/show_bug.cgi?id=2072679 - supporting progressive AVIF rendering in Firefox. This change is needed to expose the progressive layers.

Forgive me, for this was largely vibe coded, although I've manually tested it integrated with Firefox as part of https://phabricator.services.mozilla.com/D326564.

Here's some slop explaining what's going on here:

---

A layered (progressive) AVIF image item holds up to four AV1 frames concatenated in its payload, each a better rendering of the same picture. `a1lx` gives the byte sizes of the first three, and is the only signal saying where one layer ends and the next begins. It was previously recorded as an unsupported feature and skipped without reading its payload, so a caller had no way to find a layer boundary and could only decode the item whole.

Parse it, and give callers what they need to slice the payload themselves:

- `ItemProperty::LayeredImageIndexing` carries the layer sizes, and `Feature::A1lx` becomes supported.
- `AvifItem` retains its `iloc` extents and construction method, reached through `primary_item_extents`, `alpha_item_extents` and the `*_is_file_construction` predicates. A caller decoding an item incrementally needs to know which of its bytes have arrived, which the existing item-data copy cannot tell it -- and for a multi-extent item that copy would have captured the whole payload anyway.
- `primary_item_a1lx`, `alpha_item_a1lx` and `primary_item_lsel` report the properties that decide whether an item is progressively renderable. An alpha auxiliary item can be layered too, with its own `a1lx`.

A malformed `a1lx` -- truncated, overlong, or with non-zero reserved bits -- is deliberately not fatal. It is recorded as present with no layer sizes, which reads as "not a layered image". Such a file decoded fine when the property was skipped, and failing it now would be a regression.

The C API gains the matching `Mp4parseAvifInfo` fields, plus `Mp4parseItemExtents` as a borrowed slice of `ItemExtent`, valid for the lifetime of the parser.

Because `a1lx` is no longer an unsupported feature, the files whose only unsupported property it was now parse with an empty `unsupported_features` set, so they come off `AVIF_UNSUPPORTED_IMAGES`. `animals_00_multilayer_grid_a1lx` and `quebec_3layer_op2` stay, since `grid` and `a1op` are still unsupported.


mp4parse/src/lib.rs
```diff
@@ -1624,6 +1625,43 @@ impl fmt::Debug for IsobmffItem {
...
+/// which need to know *where* an item's bytes live (rather than just reading
+/// them) can be told. `extent_length` may be zero, which per § 8.11.3.3 means
+/// the extent runs to the end of the enclosing box; that is reported as
+/// `len == 0` with `to_end == true`.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct ItemExtent {
+    pub offset: u64,
+    pub len: u64,
+    pub to_end: bool,
```
baumanj: Why put the repr-C-friendly form of this in the pure-Rust part of things? The advantage to the non-repr-C-friendly `Extent` enum is that it makes the representation of illegal states impossible as well as the interpretation of the data clear from the code rather than comments. Is there an essential reason why the transformation can't be constrained to the part of the code concerned with FFI?

jakearchibald: Fair. I've moved it. Again, I tested it with the Firefox patch.

baumanj: I don't have time to review all of this closely enough to give a general approval, but with this revision, I don't see anything objectionable from a cursory perusal. Thanks for taking the time to add this cool feature.

padenot: Hi Jon hope you're well! And thanks for giving it a look. Myself or @kinetiknz will look more closely.

baumanj: Hi Paul! I miss y'all. Feel free to ping me sometime; it would be lovely to catch up.

mp4parse/src/lib.rs
```diff
@@ -3717,13 +3846,27 @@ fn read_ipco<T: Read>(
...
+                // but with no layer sizes, which callers read as "not a layered
+                // image". An `a1lx` that is truncated, overlong, or has garbage
+                // in its reserved bits is no reason to fail an image which
+                // would otherwise decode: the property "shall not be marked as
+                // essential" per
+                // <https://aomediacodec.github.io/av1-avif/#layered-image-indexing-property-description>,
+                // so a reader is entitled to ignore it entirely and still
+                // render the item -- it only ever documents where the layers
+                // are, never how to decode them.
+                let a1lx = read_a1lx(&mut b).unwrap_or_default();
```
kinetiknz: This approach swallows I/O errors silently.  I'd prefer `read_a1lx` to handle malformed entries internally.

mp4parse/src/lib.rs
```diff
@@ -3758,6 +3901,72 @@ fn read_lsel<T: Read>(src: &mut BMFFBox<T>) -> Result<u16> {
...
+    // unsigned int(7) reserved = 0; warned about rather than rejected, since
+    // the property is non-essential and this is the only field we would be
+    // guessing about.
+    if reserved_and_large_size & 0xfe != 0 {
+        warn!("a1lx reserved bits are not zero: {reserved_and_large_size:#x}");
+    }
+    // unsigned int(1) large_size; FieldLength = (large_size + 1) * 16
+    let large_size = reserved_and_large_size & 1 == 1;
+
+    let mut layer_sizes = [0u32; 3];
```
kinetiknz, line 4009: AVIF §2.3.2.3.4 says that after a zero layer size, “following values shall be 0.” Could we validate that here and ignore the index if it fails? Currently [0, 10, 0] is returned unchanged.

mp4parse/src/lib.rs
```diff
@@ -1707,6 +1725,83 @@ impl AvifContext {
...
+    pub fn alpha_item_a1lx(&self) -> Option<&AV1LayeredImageIndexing> {
+        self.item_a1lx(self.alpha_item.as_ref()?)
+    }
+
+    fn item_a1lx(&self, item: &AvifItem) -> Option<&AV1LayeredImageIndexing> {
+        match self
+            .item_properties
+            .get(item.id, BoxType::AV1LayeredImageIndexingProperty)
+        {
+            Ok(Some(ItemProperty::LayeredImageIndexing(a1lx))) => Some(a1lx),
```
kinetiknz: Before returning this index, should we check that the sum of its sizes is smaller than self.item_as_slice(item).len()? Otherwise the documented layers can consume or exceed the entire payload, leaving no final layer. This needs checking per associated item, since the same property can be shared by items with different payload lengths.

mp4parse/src/lib.rs
```diff
@@ -1707,6 +1725,83 @@ impl AvifContext {
             .map(|item| self.image_bits_per_channel(item.id))
     }
 
+    /// The layer sizes from the item's AV1LayeredImageIndexingProperty, or
+    /// `None` if the item has no `a1lx`.
+    ///
+    /// An `a1lx` "should not be associated with AV1 Image Items consisting of
```
kinetiknz, line 1740: Should this say that `Some` indicates a present property, which may contain all-zero sizes after malformed-input fallback? As the updated C documentation explains, presence alone doesn’t establish usable layer boundaries.

jakearchibald:
@kinetiknz those issues should be fixed now, and I've tested it against the Firefox patch.

kinetiknz:
That's great, thanks for the patch!

